PR #23, opened 2019-09-11
Organization API

lil2400:
Created new endpoint for organization table


zkwsk (changes requested):

zkwsk (changes requested):
Found some things I would like changed.

Furthermore, I cannot seem to get `/api/organizations/ :POST` to work. It gives me a 404. I've sent you my request in Postman on Slack. Maybe there is something I'm missing?

bytewiz (approved):
Looks good to me, could find somethings to improve but let's save that for later :v:

src/server/api/controllers/organization.controller.js
```diff
@@ -0,0 +1,88 @@
...
+    .select("*")
+    .where({
+      name
+    });
+  if (organization.length !== 0) {
+    throw new HttpError("Bad request", "organization already exists!", 409);
+  }
+  return knex("organizations").insert({
+    name: name,
+    logo: logo
```
zkwsk, line 19: Just a heads up. You have a few places where you are doing assignments like this. Could be optimized with destructuring like this:
```
{ name, logo }
```
But it's a really minor improvement.

zkwsk (approved):
Looks very good. Great work 💪

PR #25, opened 2019-09-12
Api/ports

NayanaKamtekar:
The endpoint for getting ports by id, getting ports by waypoint id and creating port are done. I have checked with dummy data. Please review.


bytewiz (changes requested):
Looks good just a few minor comments ✌️

zkwsk (changes requested):

PR #26, opened 2019-09-13
Api/ports

zkwsk:


src/server/api/routes/port.router.js
```diff
@@ -0,0 +1,54 @@
...
+    portController
+      .createPort({
+        body: req.body
+      })
+      .then(result => res.json(result))
+      .catch(next);
+  }
+);
+
+// ENDPOINT: /api/waypoints/:waypoint_id/ports/ :GET
```
zkwsk: The URL to this one is actually `/api/ports/:waypoint_id/ports/` since you put it in `ports.router.js` which in `router.js` is mappd to `/ports`.

I realize that this endpoint should probably have been part of the task to do waypoints, because the solution is to move this code to the `waypoints.router.js`, but then it also makes more sense to have `getPortByWaypointId()` in `waypoints.controller.js`.

In the meantime, however, Daniel and I discovered that we could improve the DB schema further by not connecting waypoints and ports, so we will be removing this soon, so you can actually disregard this part of the task altogether. Sorry for the extra work.

Please remove this section.

src/server/api/controllers/port.controller.js
```diff
@@ -0,0 +1,56 @@
...
+    return await knex("waypoints")
+      .select("ports.name")
+      .where("waypoints.id", id)
+      .join("ports", {
+        "waypoints.port_id": "ports.id"
+      });
+  } catch (err) {
+    console.error(err);
+  }
+};
```
zkwsk: Please remove the `getPortByWaypointId()` function (see other comment).

zkwsk:
@bytewiz has reviewed here, so I'm going to merge this one. https://github.com/HackYourFuture-CPH/ov-class09-fp/pull/25

PR #27, opened 2019-09-13
Vessels-api

hemagunti:
The endpoint is to GET and UPDATE the Vessels data.


src/server/api/controllers/user.controller.js
```diff
@@ -67,13 +67,22 @@ const createUser = async ({ body }) => {
       role: body.role
     });
 
+  const organization = await knex
```
zkwsk: I would very much like this one to be called `getOrganizationById`

src/server/api/controllers/user.controller.js
```diff
@@ -67,13 +67,22 @@ const createUser = async ({ body }) => {
...
+    .from("organizations")
+    .select("*")
+    .where({
+      id: 1
```
zkwsk: You can't hardcode and ID here. You need to get that from the url parameters when somebody is calling the endpoint (i.e. `/api/organizations/:id`).

See how it's done for users already: https://github.com/HackYourFuture-CPH/ov-class09-fp/blob/322b570df3e04a6e91f1cebb89f13cb71148315d/src/server/api/controllers/user.controller.js#L22

src/server/api/controllers/user.controller.js
zkwsk: But it belongs in `organization.controller.js` and Ilir has already created the same function there. You just need to pull develop to get it: https://github.com/HackYourFuture-CPH/ov-class09-fp/blob/322b570df3e04a6e91f1cebb89f13cb71148315d/src/server/api/controllers/organization.controller.js#L31

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,116 @@
+"use strict";
+
+const HttpError = require("../lib/utils/http-error");
+const knex = require("../../config/db");
+const bcrypt = require("bcrypt");
+const generatePassword = require("generate-password");
+const mailer = require("../lib/services/mailer.service");
+const { hashPassword } = require("../lib/utils/hash.password");
```
zkwsk: I see you copied the `user.controller.js` file as a template, which is fine. But you don't need any of these imports.

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,116 @@
...
+};
+
+//GET:id
+const getVesselById = async id => {
+  try {
+    const vessels = await knex("vessels")
+      .select("vessels.id as id", "organization_id", "mmsi", "name")
+      .where({ id: id });
+    if (vessels.length === 0) {
+      throw new HttpError("Bad request", `Cannot find user for ID ${id}!`, 404);
```
zkwsk: user?

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,116 @@
...
+
+const createVessel = async ({ body }) => {
+  const { id, organization_id, mmsi, name } = body;
+
+  const organization = await knex
+    .from("organizations")
+    .select("*")
+    .where({
+      id: organization_id
+    });
```
zkwsk, line 47: See here you could use that `getOrganizationById` function you were trying to make in the `user.controller.js`, but since it already exists in the develop branch, I suggest you pull that and import the `organization.controller.js` into this file and use the function from there, so you don't repeat work.

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,116 @@
...
+    mmsi: mmsi,
+    name: name
+  });
+  return vessels;
+};
+
+//PATCH
+const editVessel = async ({ body }) => {
+  const { id, organization_id, mmsi, name } = body;
+  console.log(body);
```
zkwsk: Remove `console.log()` statements before committing :)

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,116 @@
...
+  const vessel = await knex
+    .from("vessels")
+    .select("*")
+    .where({
+      name: name
+    });
+  if (vessel.length === 0) {
+    throw new HttpError(
+      "Bad request",
+      `Cannot find user with email ${name}!`,
```
zkwsk: user?

src/server/api/routes/router.js
```diff
@@ -22,7 +26,12 @@ router.use("/mail", mailRouter);
 
 router.use("/token", refreshTokenRouter);
 
+router.use("/vessels", vesselRouter);
+
 // /api/organizations
 router.use("/organizations", organizationRouter);
 
+// /api/organizations/organization_id/vessels
+router.use("/organizations/:organization_id/vessels", vesselRouter);
```
zkwsk: You actually need to put this endpoint in the `organization.router.js`. I realize I should probably have broken things up different and made this endpoint part of the organization task.

src/server/api/controllers/user.controller.js
```diff
@@ -67,13 +67,22 @@ const createUser = async ({ body }) => {
...
       role_id: role[0].id,
-      organization_id,
-      email,
+
+      organization_id: organization[0].id,
+      email: email,
+
       password: hashedPassword,
       name,
       status: false
```
zkwsk: Looking at the rest of your PR it looks like these changes in the `user.controller.js` was probably just you getting started and then you found out how to do it later. Always make sure to read your PR before starting it so you find out if you are committing something you were not intending to do. Happens to everyone :)

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,116 @@
...
+
+//PATCH
+const editVessel = async ({ body }) => {
+  const { id, organization_id, mmsi, name } = body;
+  console.log(body);
+  const vessel = await knex
+    .from("vessels")
+    .select("*")
+    .where({
+      name: name
```
zkwsk: You need to select by `ID`.

Right now you try to find vessels by matching the name that is passed into the parameters with what you have in the DB. But if I make a PATCH request trying to rename the vessel those two will not match.

So change this to ID and test it by sending a PATCH request with a body like this:
```
{
	"id": 1,
	"organization_id": 1,
	"mmsi": "414223",
	"name": "Fancy new name"
}
```

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return vessels;
+  } catch (err) {
+    return err.message;
+  }
+};
+//POST
+
+const createVessel = async ({ body }) => {
+  const { id, organization_id, mmsi, name } = body;
```
zkwsk: Took me a while to get to work, but that's not your fault. Apparently we are trying to save MMSI in a field that is 8 characters with two decimal spaces. It's supposed to be 9 characters integer. I'll make sure to get this fixed :)

zkwsk (changes requested):
Hey @hemagunti 👋 I've requested quite a few changes, but I think most of them are related to you committing changes to a file that probably should have been committed, because the rest is actually really solid work that only needs some minor changes 💪

src/server/api/controllers/user.controller.js
hemagunti: modified user.controller

src/server/api/controllers/vessel.controller.js
hemagunti: Could't do this since it is showing that "cannot read id of property undefined"

src/server/api/routes/router.js
hemagunti: I have kept same as before

src/server/api/controllers/vessel.controller.js
hemagunti: find vessels by matching the name instead of id

src/server/api/controllers/vessel.controller.js
hemagunti: removed console

src/server/api/routes/router.js
zkwsk: This doesn't work. You are sending the request to `vesselRouter` but in there you need to use `/` to match it: 

```
(vessel.router.js)


//ENDPOINT: /api/organizations/organizations_id/vessels
router.get(
  "/",
  authorizeUser(ROLES.SUPER_USER, ROLES.ADMIN, ROLES.OPERATOR),
  (req, res, next) => {
    ...
  }
);
```

But there is already another rule matching `/` with a GET request further up in the file, meaning that while you thing you are calling `/api/organizations/organizations_id/vessels` you are actually getting `/api/organizations`.

As I explained earlier, you need to move this rule to `organization.router.js`. From there you can import `getVesselsByOrganizationId` from `vessels.controller.js` to perform the query you need. But you have some problems in that function as well.

src/server/api/controllers/vessel.controller.js
zkwsk: You can remove `id` from that destructuring statement, since you are newer using it and it cause potential to confuse it with `organization_id`. When you create a new record you don't know the ID. The DB will provide that for you.

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,118 @@
...
+  if (organization.length === 0) {
+    throw new HttpError(
+      "Bad request",
+      `${organization_id} organization not found!`,
+      404
+    );
+  }
+
+  const vessels = await knex("vessels").insert({
+    organization_id: organization[0].id,
```
zkwsk: It seems strange and convoluted to use `organization[0].id` when you also has a variable called `organization_id` that contains the same value.

Why would you use the ID of the record you just fetched from the DB by matching it with the ID in your request if you could use the ID from your request?

If you use the ID from the request, the object you are inserting with knex looks like this:
```
{
  organization_id: organization_id,
  mmsi: mmsi,
  name, name
}
```
But since both the keys and the names of the variables you use as values are the same, why not use destructuring?
```
{ organization_id, mmsi, name }
```

Same output, more concise and easier to read :)

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,118 @@
...
+  if (Object.keys(queryVessel).length !== 0) {
+    return knex("vessels")
+      .where({ id: vessel[0].id })
+      .update(queryVessel);
+  } else return "No edit fields passed, nothing updated!";
+};
+
+//Get vessels by organization id
+const getVesselsByOrganizationId = async organization_id => {
+  const vessel = await knex()
```
zkwsk: You are not supposed to call `knex()`, but `knex.from()`. Remove `()` from this line.

src/server/api/routes/vessel.router.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+);
+
+//ENDPOINT: /api/organizations/organizations_id/vessels
+router.get(
+  "/",
+  authorizeUser(ROLES.SUPER_USER, ROLES.ADMIN, ROLES.OPERATOR),
+  (req, res, next) => {
+    vesselController
+      .getVesselsByOrganizationId(req)
```
zkwsk: You are feeding in the entire request object `req` but inside `getVesselsByOrganizationId()` you are expecting it to be an `organization_id`.

zkwsk (changes requested):

src/server/api/controllers/vessel.controller.js
bytewiz: Agree, and avoid accessing array object properties using `[0].id` it can lead to headaches later

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,118 @@
...
+    mmsi: mmsi,
+    name: name
+  });
+  return vessels;
+};
+
+//PATCH
+const editVessel = async ({ body }) => {
+  const { id, organization_id, mmsi, name } = body;
+  const vessel = await knex
```
bytewiz, line 68: Also to make it more clear that you are expecting a single result from the query you should be able to do the following `knex.from("vessels").select("*").where({name}).first()` (using `first()`) the way I see it from the knex docs

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  const queryVessel = {};
+
+  if (body.name) queryVessel.name = name;
+  if (body.mmsi) queryVessel.mmsi = mmsi;
+
+  if (Object.keys(queryVessel).length !== 0) {
+    return knex("vessels")
+      .where({ id: vessel[0].id })
```
bytewiz: From the previous comment then you can avoid using`[0].id`

bytewiz (changes requested):
Got work, but got a few comments :)

src/server/api/controllers/vessel.controller.js
hemagunti: Removed parentheses

src/server/api/controllers/vessel.controller.js
hemagunti: Used restructuring to read easy

src/server/api/controllers/vessel.controller.js
hemagunti: removed id from create vessel

src/server/api/routes/router.js
hemagunti: Removed the /api/organizations/organizations_id/vessels endpoint from vessel.router to organisation.router

src/server/api/routes/vessel.router.js
hemagunti: modified with organisation_id

zkwsk:
I think I've resolved all of the problems in this branch. I understand @hemagunti that you were working together with @Swapna-Sahu to resolve this PR. You can go to my branch and then do a "compare" with this branch once you've pushed the latest changes to see if I've solved something in a different way.

https://github.com/HackYourFuture-CPH/ov-class09-fp/tree/feature/vessels-fix-by-zaki

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,121 @@
...
+    return vessels;
+  } catch (err) {
+    return err.message;
+  }
+};
+//POST
+
+const createVessel = async ({ body }) => {
+  const { organization_id, mmsi, name } = body;
+  console.log(body);
```
zkwsk: Please get rid of the `console.log()`

zkwsk (changes requested):
Just a single console statement left ;)

src/server/api/controllers/vessel.controller.js
```diff
@@ -0,0 +1,118 @@
...
+//GET
+const getVessels = async req => {
+  try {
+    return await knex("vessels").select(
+      "vessels.id",
+      "vessels.organization_id",
+      "vessels.mmsi",
+      "vessels.name"
+    );
+  } catch (err) {}
```
zkwsk: Please return the error message back to the API, so whoever is consuming the API has a chance of understanding what went wrong.
```
} catch (err) {
  return err.message;
}
```

zkwsk (changes requested):

src/server/api/controllers/vessel.controller.js
hemagunti: removed console

src/server/api/controllers/vessel.controller.js
hemagunti: fixed it.

zkwsk (approved):

bytewiz (approved):

PR #29, opened 2019-09-14
Selected routes composite key

zkwsk:
We had some missing primary key constraints in the DB. A couple of tables needed composite primary keys (primary key consisting of multiple fields) and since we could not work out how to model that in dbdiagram, we forgot to add it. That's fixed now.


zkwsk:
Helping Sheila and it seems she needs this in her branch, so I'm going to use god mode and approve it myself :D

PR #31, opened 2019-09-14
Voyages api

ghost:
Voyages API is created and tested!

Note: if you want to post a voyage before that be sure you have at least one row data in the vessels table.

in case you need to make a new post in the Postman you can copy, paste...

{
    "vessel_id": 1,
    "etd": "2019-09-10 21:20:09",
    "eta": "2019-09-10 21:20:09",
    "forwarddraft": 8.596,
    "aftdraft": 8.785,
    "optimisation_type": 9.852,
    "depart_from": 8,
    "arrive_at": 7,
    "sailtime": 98.879,
    "hfocost": 89.63,
    "lsfocost": 87.52,
    "hirerate": 963.87,
    "status": "created"
}

### DIFF
--- a/src/server/api/controllers/voyages.controller.js
+++ b/src/server/api/controllers/voyages.controller.js
@@ -1,0 +1,37 @@
+"use strict";
+
+const HttpError = require("../lib/utils/http-error");
+const knex = require("../../config/db");
+
+// to get all the voyages
+const getVoyages = () => {
+  return knex("voyages").select("*");
+};
+
+//to create a voyage
+const createVoyage = async ({ body }) => {
+  try {
+    return knex("voyages").insert({
+      vessel_id: body.vessel_id,
+      etd: body.etd,
+      eta: body.eta,
+      forwarddraft: body.forwarddraft,
+      aftdraft: body.aftdraft,
+      optimisation_type: body.optimisation_type,
+      depart_from: body.depart_from,
+      arrive_at: body.arrive_at,
+      sailtime: body.sailtime,
+      hfocost: body.hfocost,
+      lsfocost: body.lsfocost,
+      hirerate: body.hirerate,
+      status: body.status
+    });
+  } catch {
+    console.log("something is wrong!");
+  }
+};
+
+module.exports = {
+  getVoyages,
+  createVoyage
+};
--- a/src/server/api/controllers/voyages.controller.js
+++ b/src/server/api/controllers/voyages.controller.js
@@ -27,7 +27,7 @@ const createVoyage = async ({ body }) => {
       status: body.status
     });
   } catch {
-    console.log("something is wrong!");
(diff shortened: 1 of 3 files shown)


src/server/api/controllers/voyages.controller.js
```diff
@@ -0,0 +1,37 @@
...
+      depart_from: body.depart_from,
+      arrive_at: body.arrive_at,
+      sailtime: body.sailtime,
+      hfocost: body.hfocost,
+      lsfocost: body.lsfocost,
+      hirerate: body.hirerate,
+      status: body.status
+    });
+  } catch {
+    console.log("something is wrong!");
```
zkwsk: Please use a more descriptive text, and preferably use `console.error()`.

For example:
`console.log("Error: Could not insert voyage.")`

src/server/api/routes/voyages.router.js
```diff
@@ -0,0 +1,34 @@
...
+  (req, res, next) => {
+    voyagesController
+      .getVoyages()
+      .then(result => res.json(result))
+      .catch(next);
+  }
+);
+
+// ENDPOINT: /api/voyages/ :post
+router.post("/", authorizeUser(ROLES.SUPER_USER), (req, res, next) => {
```
zkwsk: This one should be accessible to superuser, admins and operators. Currently, only superusers can use it.

src/server/api/routes/voyages.router.js
```diff
@@ -0,0 +1,34 @@
...
+
+const ROLES = require("../../constants/roles");
+
+// controllers
+const voyagesController = require("../controllers/voyages.controller");
+
+// ENDPOINT: /api/voyages/ :GET
+router.get(
+  "/",
+  authorizeUser(ROLES.SUPER_USER, ROLES.ADMIN),
```
zkwsk: This one should only be accessible to superusers.

Reason being that this is the endpoint to get all voyages in the entire application. It's really helpful to have, but it's not something that the operators or admins will ever have access to, since it would show data from outside their own organization. So this endpoint is mainly for debugging and maybe in the future to enable doing reports or something like that for Optimum Voyage. It's not for the end users.

End users will instead use the "Get voyages by vessel" endpoint.

zkwsk (changes requested):

zkwsk:
Also, where is the "Get voyages by vessel" endpoint? That's actually the most important one.

I know it's a little confusing, but you're actually supposed to make that in the `vessels` router, which Hema is working on (see her PR).

If you start working in that file you will almost certainly get merge conflicts, so it might be a good idea to merge this on first, then branch out from her branch and make the last endpoint.

Another option is to have a look at what she did on her branch, create the files with the same names and structure (not with her endpoints) and add your endpoint and solve any merge conflicts when they come. You can't avoid merge conflicts on a project like this.

PR #33, opened 2019-09-15
Added json-server to package.json

lil2400:
Added json-server for FAKE REST API


zkwsk (changes requested):
Why does git show a diff on a whole bunch of changes that look exactly like what you already merged here? Don't like this 😟 

https://github.com/HackYourFuture-CPH/ov-class09-fp/pull/23/files

bytewiz (changes requested):
Agree, can we remove the parts that are already there? this should not include all the organization part

lil2400:
I might have done a mistake. I have not done git pull before commiting

zkwsk (dismissed):
Now it looks good.

zkwsk:
Hej @lil2400 :) I don't actually think you want to have a pull request going for this one? Looks like you're still developing on it, or are you done?

zkwsk:
Oh, I realize now that is actually working. But I would like some changes. Review coming up :)

package.json
```diff
@@ -11,7 +11,8 @@
     "start": "npm run build && node ./src/server/bin/server",
     "client": "webpack-dev-server --mode development --devtool inline-source-map --hot",
     "server": "nodemon ./src/server/bin/server",
-    "dev": "concurrently \"npm run server\" \"npm run client\"",
+    "json": "cd ./src/external-api/ && node server.js",
+    "dev": "concurrently \"npm run server\" \"npm run client\" \"npm run json\"",
```
zkwsk: Please call the task `json-server` instead of just `json`. Or you could call it `external-api` if you prefer.

src/external-api/data/vessel_list.json
```diff
@@ -0,0 +1,24 @@
+{
+  "vessel_list": [
+    {
+      "id": "1",
+      "name": "SS_Martha",
+      "company_id": "0"
```
zkwsk: I just noticed that we do not have any information about the companies. Can you please create another json file with some fake company names (no need to ask Optimum Voyage to output real customers and then obscure the names after).

I guess we only need 4 company names/ids with the current data, but if you want to go down the easy route you could use https://mockaroo.com/ to generate some fake data :)

src/external-api/data/vessel_list.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    {
+      "id": "4",
+      "name": "Seven_Seas",
+      "company_id": "1"
+    },
+    {
+      "id": "5",
+      "name": "MS_Marie",
```
zkwsk: Also just realized that the names have been snake_cased. That looks a bit silly in the UI. Maybe just change it to "MS Marie" and so on...?

src/external-api/index.js
```diff
@@ -0,0 +1,13 @@
+const routeInfo = require("./data/route_info_dump.json");
+const vesselList = require("./data/vessel_list.json");
+const vesselReport = require("./data/vessel_report_dump.json");
+const voyageInfo = require("./data/voyage_info.json");
+
+module.exports = function() {
+  return {
+    route_info: routeInfo,
+    vassel_list: vesselList,
+    vesselreport: vesselReport,
```
zkwsk: Could we call it `vessel_report` for consistency?

src/external-api/index.js
```diff
@@ -0,0 +1,13 @@
...
+    voyageinfo: voyageInfo
```
zkwsk: And `voyage_info`, please? :)

src/external-api/server.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+const port = 4000;
```
zkwsk: Throw this guy in an env variable: `process.env.external_api_port`.

And add that env variable to the `.env.example` file.

(and to your local `.env` file to make it work, obviously)

zkwsk (changes requested):
Really great work @lil2400. I have requested some changes, but I realize that if you should have gotten that right the first you would probably have had to read my mind 😆

lil2400:
Hi Zaki, no I have not finished. I did make pull because I wasn't sure if the task I did was good, I wanted review

src/external-api/data/vessel_report_dump.json
```diff
@@ -0,0 +1,17489 @@
...
+        "lon":null,
+        "lsfo_con":0.0,
+        "date":"DMY",
+        "report_type":"Report Type",
+        "time":"UTC",
+        "lat":null,
+        "hfo_con":0.0,
+        "speed":"SPD",
+        "vessel_id":1
+     },
```
zkwsk: The first record here is a mistake on Thors side. Delete that.

zkwsk (changes requested):
Overall the approach is right, but there is something that we have not really discussed.

You need to have linked data, meaning that when you get a `vessel_report` you need to be able to retrieve the voyage and from the voyage you need to retrieve the `suggested_routes` and so on. How do you imagine doing that?

I have a couple of ideas.

1. You could make a script that loads the different json files and builds a js object with all of the data linked. Then you could export the JS object to JSON using the JSON.stringify() method and either print it to the console or to a file. If you print to the console make sure to have history long enough that you can copy everything into a file after. Now use the new file that looks something like this:

```
{ 
  "vessel_reports":{ 
    "report_id":1,
    "RPM":"",
    "lon":null,
    "lsfo_con":1.2,
    "date":"2019-09-23::20:59:12",
    "report_type":"DEPARTURE",
    "time":"",
    "lat":null,
    "hfo_con":0.0,
    "speed":"",
    "vessel":{ 
      "id":"1",
      "name":"SS_Martha",
      "company":{ 
        "id":1,
        "name":"Maersk"
      }
    }
  }
}
```

You would probably have to attach the voyage the vessel is on and a bunch of other things, but the idea is that you just get one major object with a dump of all data for each vessel_report. I'm not sure that's the best way though, because think about this endpoint be called maybe every 10 seconds. It's a lot of data to transfer.

Another approach could be to make more endpoints on JSON server and when you are populating our DB from the `vessel_reports`you first check if the report is already "known" to our system. If it is, you can throw it away and do nothing. If it is not you start calling all of the other endpoints to get the vessel, the voyage and so and so on and check if each of them needs to be updated in our system. This would allow us to not fetch all data every time, but on the other hand we would need to make a lot more requests (given there is a new vessel_report). I'm not entirely sure which approach is best. But it really does not have to be perfect. We are just trying to "mock" a fake api right now, so which solution do you think is best?

zkwsk:
I'm going to close this one, @lil2400. But let's keep the branch around in case we want to repurpose some of the work you did :)

PR #37, opened 2019-09-17
routes router & controller has been added , router file has been updated

AbodHassam:


zkwsk (changes requested):
Hey @AbodHassam 👋 Nice to see your pull request online.

I have a quite critical change you need to address before I can go more in depth with your PR. Problem is that since you started this PR we have renamed `routes` to `suggested_routes` in the DB.

You are consistently calling the `routes` table in you SQL, meaning that you clearly have not updated your db by running the latest migrations.

What you need to do is the following:
- Run either `knex migrate:latest` (if you've installed knex locally) or alternatively `npm run db:setup` to get your DB up to date. Now the `routes` tables should be called `suggested_routes` in your DB and your code will break.
- Update all references to `routes` in your SQL (knex calls) to `suggested_routes`.
- Test that it works again?

Furthermore, I would like you to refactor any references to routes in your code, so the file `routes.router.js` needs to be renamed, the endpoint we are calling needs to be renamed to `api/suggested_routes/` and any variable names containing `routes` should be changed to `suggested_routes`. You don't need to change foreign key names in the database, though, we have another task for that. If you feel like this task is overwhelming reach out and we can discuss merging the PR without this and make a separate task for refactoring (the most important part is to get it to work with the DB changes).

Sorry about this change. I've updated your trello task accordingly. This is a very good reason why it is important to follow along and read the pull requests that get merged here: https://github.com/HackYourFuture-CPH/ov-class09-fp/pulls?q=is%3Apr+is%3Aclosed

Finally, you are still missing the endpoint "Get routes by voyage id". It's a really important endpoint, but I realize it is kind of difficult to do, so let's get this PR merged without that task first and then do the last endpoint in a separate PR?

zkwsk (dismissed):
Really good work ⛱ 👍 🥇 

I split out the last endpoint in a new task here: https://trello.com/c/aAClnsUo/36-api-endpoint-get-routes-by-voyage-id

You need one more review before we can merge, though.

src/server/api/controllers/routes.controller.js
```diff
@@ -0,0 +1,64 @@
...
+    distance_through_water,
+    avgspeed
+  } = body;
+  const route = await knex
+    .from("routes")
+    .select("*")
+    .where({
+      id: voyage_id
+    });
+  if (route.length !== 0) {
```
ismailirahim: Hi Abod, here I think it should be === instead !==

zkwsk: No it's not. But you're right that there is a problem here. We cannot create more than one suggested route that is connected to a voyage and we need to be able to do that.

zkwsk (changes requested):
@AbodHassam you need to remove the check from line :28 to :36 in `suggested_routes.controller.js`. The check is blocking you from creating more suggested_routes with the same voyage. But the point is we need many suggested_routes for each voyage, so the shipping operators can decide which one they want to take.

For some reason I cannot comment inline in code after dismissing my preview review.

zkwsk:
Somehow we got a bunch of commits that don't belong on this branch. I'm trying to fix it on another branch.

PR #39, opened 2019-09-18
Datetime columns migrations

NayanaKamtekar:
I have done migrations to change columns ETA and ETD in table VOYAGES, and column ETA in table SUGGESTED_ROUTES to datetime.


zkwsk (approved):
Really solid work @NayanaKamtekar. I tested both the up and down migration.

If anyone wants to test this locally while reviewing, make sure to go to the `/src/server` folder and run `knex migrate:latest`. Check that the etd and eta fields in the database changes to datetime where it was date before. After reviewing, make sure to run `knex migrate:rollback` so your DB is not in a "future state" before Nayanas branch is actually merged into develop.

bytewiz (approved):

PR #44, opened 2019-09-20
Create seeds

fawadik:
Test users created


zkwsk (dismissed):
Looks good 👍

src/server/seeds/dev/005_00_ports.js
```diff
@@ -0,0 +1,22 @@
...
+          id: 2,
+          name: "port2"
+        },
+        {
+          id: 3,
+          name: "port3"
+        }
+      ]);
+    });
+};
```
zkwsk, line 22: Please give some more realistic names to the ports and add some more.

Can you add the ports from this list and link a waypoint to each of them with the coordinates listed in the json file?

https://trello-attachments.s3.amazonaws.com/5d741a742ea45775126dc294/5d74f07f20624e860839e985/74258c233230498015ce407a2a4ff688/harbour_info.json

src/server/seeds/dev/006_00_vessels.js
```diff
@@ -0,0 +1,28 @@
...
+          name: "hyf-vessel2"
+        },
+        {
+          id: 3,
+          organization_id: 1,
+          mmsi: 24.33,
+          name: "hyf-vessel3"
+        }
+      ]);
+    });
```
zkwsk, line 27: Please add more realistic names to the vessels and add some more. You can use something like Mockaroo to generate fake data. Don't know if it does ship names :D

zkwsk: Also, could you add some more organizations (again use Mockaroo - you can definitely get company names there) and then spread the vessels over all of the organizations.

src/server/seeds/dev/008_00_voyages.js
```diff
@@ -0,0 +1,58 @@
...
+  return knex("voyages")
+    .del()
+    .then(function() {
+      // Inserts seed entries
+      return knex("voyages").insert([
+        {
+          id: 1,
+          vessel_id: 1,
+          etd: "2019-12-20",
+          eta: "2018-11-23",
```
zkwsk, line 12: Would be nice with some more voyages along with more vessels. Make sure that they don't all have the same etd and etas like now. And make sure that etd is always before eta (you can't arrive before you depart).

src/server/seeds/dev/010_00_suggested_routes.js
```diff
@@ -0,0 +1,46 @@
...
+          id: 3,
+          voyage_id: 3,
+          eta: 300,
+          max_wave_height: 100,
+          hfo: 10.12,
+          lsfo: 123.56,
+          total_cost: 23456,
+          distance_over_ground: 2345,
+          distance_through_water: 234234,
+          avgspeed: 234.45
```
zkwsk, line 42: With more vessels and voyages we also need more routes.

src/server/seeds/dev/012_00_waypoints.js
```diff
@@ -0,0 +1,31 @@
...
+          lon: 2344.44,
+          lat: 2234.87
+        },
+        {
+          id: 3,
+          route_id: 3,
+          port_id: 3,
+          lon: 2344.44,
+          lat: 2234.87
+        }
```
zkwsk, line 28: We need waypoints associated with all of the routes. And we also need waypoints for the ports, but use the json file I linked above for that.

Also note, that route waypoints will have port_id set to null and port waypoints will have route_id set to null.

zkwsk (approved):
It works now, but I would like to have better quality data. I've seen many times that there are bugs in the application because there is either not enough data to show what the application look like in reality, or the names are so generic that you have no idea whether you are looking at a vessel or a route or something else. So it's worth putting a little effort into.

But since we might need this for demos tomorrow, I suggest we get it merged now and you add the improvements I suggested here to the same branch after.

zkwsk:
Merged this one early as it will be nice to have before API demos and it is working now. But @fawadik will continue resolving my change requests on the same branch.

PR #41, opened 2019-09-19
Voyages column names change

NayanaKamtekar:
I changed column names as a part of task: Renaming inconsistencies for arrive_at_waypoint, depart_from_waypoint .

PR #51, opened 2019-09-21
Voyages foreign keys

NayanaKamtekar:
I have changed the columns `depart_from_waypoint` and `arrive_at_waypoint` to `unsigned`, and then created them as foreign keys.
Please note, all the migrations with voyages_ including one from the branch `voyages-columns-name-change` (pull request #41) should be run together in sequence as there are dependencies.


src/server/migrations/20190921215818_voyages_changed_type_of_foreign_keys.js
```diff
@@ -0,0 +1,21 @@
+"use strict";
+
+exports.up = function(knex) {
+  return knex.schema.alterTable("voyages", table => {
```
zkwsk, line 4: Notice that the `.alterTables()` is used instead of `.tables()`.

zkwsk (commented):
Looks good, but I need to test locally before approving.

zkwsk (approved):

bytewiz (approved):
Looks good to me 👍

PR #53, opened 2019-09-22
Waypoints endpoint 

ismailirahim:
**Create waypoint:**
`/api/waypoints/ :POST`

**Get a waypoint:**
`/api/waypoints/:id :GET`

Sub resources
---

**Get all waypoints for a single route:**
`/api/routes/:route_id/waypoints/ :GET`


src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,82 @@
...
+  } catch (err) {
+    return err.message;
+  }
+};
+
+// Method for creating a waypoints
+
+const createWaypoints = async ({ body }) => {
+  const { lat, lon, port_id, route_id } = body;
+  console.log("this is body", body);
```
lil2400: Remove console

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,82 @@
...
+  // 1. Get the route
+  const routes = await knex
+    .from("suggested_routes")
+    .select("*")
+    .where({ id: route_id });
+  if (routes.length === 0) {
+    throw new HttpError(
+      "Bad request",
+      `Cannot find routes for ID ${route_id}!`,
+      409
```
lil2400: Shouldn't the error be 404? Instead of 409!

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+  }
+  console.log("routes", routes);
```
lil2400: Remove console.log

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,82 @@
...
+  // 2. Get the port
+  const ports = await knex
+    .from("ports")
+    .select("*")
+    .where({ id: port_id });
+  if (ports.length === 0) {
+    throw new HttpError(
+      "Bad request",
+      `Cannot find ports for ID ${port_id}!`,
+      409
```
lil2400: Here 404 as well!

src/server/api/controllers/waypoints.controller.js
zkwsk: Yeah, 404 here and same further down.

zkwsk (changes requested):
~~As far as I understand what you've made, there is something wrong with the logic of it.~~

~~As far as I can understand you are assuming that there is already a suggested route that needs to have a series of waypoints attached to it. What I was expecting was that the endpoint should _create_ a new suggested route and right after creating that suggested route it should attach a number of waypoints to it.~~

~~In practice we never create suggested routes without having waypoints associated with them, so this would actually be the endpoint that we would be using to create new suggested routes in practice.~~

I'm sorry, I misunderstood which task it was. Don't mind this comment.

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,81 @@
...
+    .from("ports")
+    .select("*")
+    .where({ id: port_id });
+  if (ports.length === 0) {
+    throw new HttpError(
+      "Bad request",
+      `Cannot find ports for ID ${port_id}!`,
+      404
+    );
+  }
```
zkwsk, line 40: We sort of discussed this already, but as soon as we get the port set to be nullable, we just set null on the port and this DB request is pointless :)

zkwsk:
Never mind my previous comment about creating routes with waypoints. I thought it was another task :D

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,81 @@
...
+    return waypoints;
+  } catch (err) {
+    return err.message;
+  }
+};
+
+// Method for creating a waypoints
+
+const createWaypoints = async ({ body }) => {
+  const { lat, lon, port_id, route_id } = body;
```
zkwsk: `route_id` is already renamed to `suggested_route_id` on develop. You need to pull develop and resolve merge conflicts. I recommend that you do it after my PR fixing the seeds has been merged.

zkwsk (changes requested):

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,66 @@
...
+  });
+};
+
+// Method for getting waypoints by route id
+const getWaypointsForRoute = async id => {
+  try {
+    return await knex("suggested_routes")
+      .select("*")
+      .where({ suggested_route_id: id })
+      .join("waypoints", { suggested_route_id: "suggested_routes.id" });
```
zkwsk: You are joining suggested route and waypoints and selecting everything, so we get all of the metadata for suggested routes on every waypoint. Please make your select more specific.

zkwsk (changes requested):

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,66 @@
+const HttpError = require("../lib/utils/http-error");
+const knex = require("../../config/db");
+
+// Method for getting waypoints by id:
+const getWaypointsById = async id => {
+  try {
+    const waypoints = await knex("waypoints")
+      .select("id", "lon", "lat", "created_at", "updated_at")
```
zkwsk: I know your PR was first in the line, but I needed to get the DB changes in there first, so unfortunately these have changed names to "longitude" and "latitude" respectively.

zkwsk: Make sure to check the rest of the PR for the same changes.

zkwsk (changes requested):

zkwsk (approved):
Good job! 👍 We need to get the swagger documentation updated, but let's get this one merged and do that in a separate PR.

PR #55, opened 2019-09-22
Vessel reports.position waypoint

ghost:
first I rename "position" column in "vessel_resports" table to "position_waypoint"
then I change "position_waypoint" type to int(10)unsigned
finally I made "position_waypoint" column as the foreign key from "waypoints.id"

NOTE: you have to delete "009_00_vessel_reports.js" in SEEDS folder than you have to make a new seed.


zkwsk (approved):

zkwsk:
I'm just going to rush this through as the seed files are broken after some new db migrations were introduced.

PR #58, opened 2019-09-23
Updated Schema.dbml and Schema.png

NayanaKamtekar:
I have updated Schema.dbml and Schema.png.


zkwsk (changes requested):
You have conflicts.

Also, make sure to copy all of the dbml to dbdiagram.io, save, copy the url and update the link in the readme file, so we have the latest version linked.

zkwsk (approved):
There are still some inconsistencies, but they are old and let's just try to get it ironed out next time. Stuff like `fk_user_id` and such.

PR #61, opened 2019-09-23
Bug migration waypoints

ismailirahim:
Hey @zkwsk can you have a look please


zkwsk (changes requested):
Unfortunately, it's not quite right. You did make it optional, but you also removed the foreign key constraint. I've also tried like a 1000 ways and can't figure it out :'(

src/server/api/controllers/waypoints.controller.js
```diff
@@ -0,0 +1,81 @@
+const HttpError = require("../lib/utils/http-error");
+const knex = require("../../config/db");
+
+// Method for getting waypoints by id:
+const getWaypointsById = async id => {
+  try {
+    const waypoints = await knex("waypoints")
+      .select("id", "lon", "lat", "created_at", "updated_at")
```
zkwsk, line 8: Sorry, but I merged Omids PR renaming to "longitude" and "latitude" respectively first, so you need to pull in the latest develop and rename these.

zkwsk (changes requested):

zkwsk:
I'm a little confused right now. What exactly does this PR contribute? How is it related to #53 ?

(I think maybe you have stuff in your file diff from a different branch or maybe I've just lost track of what's going on 😆 )

zkwsk:
I'm pretty sure this is solved in #64. Please reopen if I am wrong :)

PR #62, opened 2019-09-23
Fix gps coordinates in db ;)

ghost:
ecas & waypoints, columns renamed to longitude & latitude.

then the types are changed to. 
- Latitude: DECIMAL(8,6)
- Longitude: DECIMAL(9,6)


zkwsk (changes requested):
Looks good, but you need to update everywhere where we use `lat` or `lon`. In the seed files and in the controllers when interacting with the DB especially.

ghost:
@zkwsk Please take a look at SEEDS

zkwsk (approved):
You had forgot to update the ports waypoints file, so I had to do that myself. But I'm rushing this one through, so we can try to close down the DB model changes. Always test your migrations by wiping the db and running `npm run db:setup`.

PR #64, opened 2019-09-23
Fix waypoint constraints

zkwsk:
Though one. Spent more than 2 hours figuring out how to make foreign key columns optional with knex and then another 2-3 hours fixing the seed 🙀 

Finally ended up solving the migration with `knex.raw()`, writing "pure" SQL.

Furthermore, I've fixed the seeds. There were quite a few steps involved:

1. We needed "port waypoints" to be seeded before voyages, but regular "suggested route waypoints" needed to be seeded after voyages. Solution was to split waypoints into two migration files and run the "port waypoints" separately before voyages and "suggested routes waypoints" after voyages.

2. Removed ports from suggested route waypoints and suggested routes from the ports waypoints. Not strictly necessary to run the seed, but this is the format the data should have.

3. There were some fields dates which I changed to datetime, but I think the database is lenient enough to just guess and set the time. I noticed another place where we entered "500" as a datetime, the database just set the value to sometime in 2003. So this step was probably not necessary to run the seed.

4. Now after the waypoints seed had been split into two files, both of them were still starting with a `.del()` call that would clear out the table, meaning that when waypoint file number 2 would start it would remove the work that the first file had done. Removed the `.del()` call in the 2nd file.

5. Found that the entries in both waypoint seed files was declaring primary keys and starting at 1, so they would attempt to overwrite each other. Removed primary keys from all files. They are auto-incrementing and it's better to let the database handle creating the keys so we are free to insert records where we need to.


ismailirahim (approved):
It was hell lot of work.... Looks fine to me.

zkwsk:
Merging this one early as loads of things depends on it.

PR #72, opened 2019-09-24
vessels-details

Swapna-Sahu:
Created React component for vessel report
Added fake story with fake data
created react component for route view
Add react component for route view

[TASK](https://trello.com/c/SQEuWKs7/19-user-can-see-the-organizations-vessels-on-a-list)


bytewiz (changes requested):
So looking purely at the HTML structure as I understand was the task :)

`<label>` is usually reserved to be used with `<input>` or similar elements: [check here](https://www.w3schools.com/tags/tag_label.asp)

I would consider switching to `<p>` instead and generally just stay to the hierarchy of `<h1,2,3,4>` tags for the headlines and then `<p>` for the text

Also might just be me but the previous task you had looks like it already handles what you got from line 17-23 and should be possible to remove 👍 in my opinion

Also generally the components seem very repeative. I would consider the following:

```
import React, { Component } from "react";

function Box(title, data) {
  return (
    <div>
      <h4>{title}</h4>
      <p>{data}</p>
    </div>
  );
}

function BoxHeader({ title }) {
  return <h3>{title}</h3>;
}

function BoxRow({ children }) {
  return <div>{children}</div>;
}

function VesselsDetailsComponent({
  vesselName,
  departureFrom,
  arrivesAt,
  etd,
  eta,
  date,
  lat,
  lon,
  hsfo,
  lsfo
}) {
  return (
    <div>
      <BoxHeader title="Position" />
      <BoxRow>
        <Box title="DATE" data={date} />
        <Box title="LATTITUDE" data={lat} />
        <Box title="LONGITUDE" data={lon} />
      </BoxRow>
      <BoxHeader title="Consumption" />
      <BoxRow>
        <Box title="HSFO" data={hsfo} />
        <Box title="LSFO" data={lsfo} />
        <Box title="FOC TOTAL" data={hsfo + lsfo} />
      </BoxRow>
    </div>
  );
}

export default VesselsDetailsComponent;
```

At least I find that much easier to read (might be seem overkill with the `BoxHeader` but helps readability) let me know if you agree or not. 👍

zkwsk:
Good point about `<label>` belonging to inputs.

You are introducing Bow and BoxRow. That's based on the layouting tools in Material UI, right? I was thinking to keep both layout and styling entirely out of the UI components for now, but maybe we should just get used to using the basic layouting tools?

bytewiz:
> Good point about `<label>` belonging to inputs.
> 
> You are introducing Bow and BoxRow. That's based on the layouting tools in Material UI, right? I was thinking to keep both layout and styling entirely out of the UI components for now, but maybe we should just get used to using the basic layouting tools?

@zkwsk It is more just the structure and readability imo and I think e.g. material-ui uses the concept of `Cards` for this kind of design :)

zkwsk:
Oh. Right. Now I understand. Yeah I was thinking of these information boxes as cards, so @Swapna-Sahu let's make a Card component for them with a label and a value.

I don't like `BoxHeader` and neither `CardHeader` because I use it as the header for other things as well. I think it is just a generic `<h2>` or `<h3>` header. At least for now. I have not really put much effort into the type hierarchy yet.

zkwsk:
Just had a chat with @Swapna-Sahu. We discussed breaking `VesselDetails` into more components. Just to share what we agreed on:

This one should be called `Card`:
Screen Shot 2019-09-28 at 14 31 39

This should be called `CardGroup`:
Screen Shot 2019-09-28 at 14 32 40

No need to make the header (position) a separate component. And let's not have "row" components, because maybe I'll change the row length at some point :)

And for the top part of the component, reuse `VesselTeaser`.

Swapna-Sahu:
with ref to  Pull Request #89

PR #74, opened 2019-09-25
Supper user creating admin ;)

ghost:
by the this API: 

> /api/organizations/:organization_id/users/ :POST

you can make an Admin:
ror example:

> {
>     "name": "aUser",
>     "email": "user@gmail.com",
>     "password": "userpass"
> }


src/server/api/controllers/user.controller.js
```diff
@@ -11,9 +11,7 @@ const getUsers = async req => {
   try {
     return await knex("users")
       .select("users.email", "users.name", "user_roles.role")
-      .join("user_roles", {
-        "users.role_id": "user_roles.id"
-      });
+      .join("user_roles", { "users.role_id": "user_roles.id" });
```
lil2400: Do you have installed Prettier in the vs code? Because I beilive most of us have and you kinda changing the formats here. Which we might have problem in the future, because Prettier it does format automatically.

ghost: We are here and discussing it. and thanks for mentioning.

ghost: Please review the code and if it was something wrong let me know. ;)

bytewiz (dismissed):
Looks fine to me 👍 but yea a bit harder to check changes when a lot is just formatting

ghost:
@zkwsk

zkwsk (changes requested):
@Omid-Incarnation were you planning to change those format changes back like we discussed?

ghost:
@zkwsk take a look at this please

src/server/api/controllers/user.controller.js
```diff
@@ -295,6 +295,24 @@ const editUserRole = async ({ body, id }) => {
...
+  const users = await knex
+    .from("users")
+    .select("*")
+    .where({ email: body.email });
+  if (users.length !== 0) {
+    throw new HttpError("Bad request", "user already exists!", 409);
+  }
+  const hashedPassword = await hashPassword(body.password);
+  return knex("users").insert({
+    role_id: 2,
```
zkwsk: Instead of hardcoding the id of the admin role, please hardcode the name of the role.

Meaning, make a lookup in the db finding the user role with the name "admin", get the ID and then pass the resulting ID into the insert function. This ensures that id someone changes the order in the seed that makes the roles (or the IDs change in some other way) you don't suddenly end up creating wrong user roles.

See how we did it here: https://github.com/HackYourFuture-CPH/ov-class09-fp/blob/develop/src/server/api/controllers/user.controller.js#L74

zkwsk (changes requested):

ghost:
@zkwsk 
Please take a look at the latest change.
I test is and it was working.

Swapna-Sahu (commented):
Reviewed. All looks good other than one console.log inside deleteUser method.

ghost:
@Swapna-Sahu 
Thanks for mentioning, but I didn't add that function.
@zkwsk If you think I have to delete it please let me know.

zkwsk (approved):

PR #75, opened 2019-09-25
Route waypoints

ismailirahim:
This one extends on the routes endpoint.

You will typically not create a route without also creating at least two waypoints.

We need to alter the following endpoint:

**Create route with waypoints:**
`/api/routes/ :POST`


src/server/api/controllers/routes.controller.js
```diff
@@ -0,0 +1,58 @@
...
+    })
+    .then(function([id]) {
+      let sequence_id = 1;
+      const waypointsForRoute = waypoints.map(waypoint => {
+        return {
+          route_id: id,
+          port_id: port_id,
+          sequence_id: sequence_id++,
+          lon: waypoint.lon,
+          lat: waypoint.lat
```
zkwsk: Sorry 🙈 Same story again: we ranamed `lat` and `lon` :(

zkwsk (changes requested):

src/server/api/routes/routes.router.js
```diff
@@ -0,0 +1,19 @@
...
+const routesController = require("../controllers/routes.controller");
+
+router.post("/", (req, res, next) => {
+  routesController
+    .createArouteWithWaypoints({
+      body: req.body
+    })
+    .then(result => res.json(result))
+    .catch(next);
+});
```
zkwsk: We renamed `routes` to `suggested-routes` all around. You need to move this to this file: https://github.com/HackYourFuture-CPH/ov-class09-fp/blob/develop/src/server/api/routes/suggested-routes.router.js

src/server/api/controllers/routes.controller.js
```diff
@@ -0,0 +1,58 @@
...
+    eta,
+    max_wave_height,
+    hfo,
+    lsfo,
+    total_cost,
+    distance_over_ground,
+    distance_through_water,
+    avgspeed,
+    waypoints
+  } = body;
```
zkwsk: There is a small thing with naming conventions here. I would like to use camelCase in javascript.

In the database we use snake_case and in paths/urls and in filenames we use hypen-case (or kebab-case as some people call it?).

src/server/api/routes/routes.router.js
```diff
@@ -0,0 +1,19 @@
...
+const express = require("express");
+const router = express.Router({
+  mergeParams: true
+});
+
+const routesController = require("../controllers/routes.controller");
+
+router.post("/", (req, res, next) => {
+  routesController
+    .createArouteWithWaypoints({
```
zkwsk: There is a typo here. Should be `createRouteWithWaypoints`. Also, since we renamed routes to suggested routes, please call it `createSuggestedRouteWithWaypoints` and rename it in the controller as well.

src/server/api/controllers/routes.controller.js
```diff
@@ -0,0 +1,58 @@
+const HttpError = require("../lib/utils/http-error");
+const knex = require("../../config/db");
+const moment = require("moment-timezone");
+
+const createRouteWithWaypoints = async ({ body }) => {
```
zkwsk: Please rename to `createSuggestedRouteWithWaypoints`

zkwsk (changes requested):

zkwsk (approved):

PR #76, opened 2019-09-25
MMSI type is changed ;)

ghost:
MMSI type is changed from decimal to varchar with 9 characters.


src/server/seeds/dev/006_00_vessels.js
```diff
@@ -12,17 +12,17 @@ exports.seed = function(knex, Promise) {
       return knex("vessels").insert([
         {
           organization_id: 1,
-          mmsi: 23.33,
+          mmsi: "whatever",
```
fawadik: i am not sure if it is really required to change the seed values as well, because i have now changed this seed into some real data.....!
but i think it is fine until i send the pull request

src/server/migrations/20190925112652_vessels_MMSI_type_change.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+exports.up = function(knex) {
+  return knex.schema.alterTable("vessels", table => {
+    table
+      .string("MMSI", 9)
```
fawadik, line 6: is it okey to use "string" or "varchar", because in the task it is mentioned to change the type from decimal to varchar

src/server/seeds/dev/006_00_vessels.js
ghost: @fawadik as @zkwsk asked me before, whenever I have to do change the migrate file I have to make the proper change to the seeds.

src/server/migrations/20190925112652_vessels_MMSI_type_change.js
ghost: @fawadik In knex "string" means "varchar". if you run the code and take look at DB you can see the previous type is changed to varchar.

src/server/seeds/dev/006_00_vessels.js
ghost: MMSI - Maritime Mobile Service Identity it’s the ID that is used for radio/frequency communications and it can be used as an unique identifier of a Vessel.

src/server/migrations/20190925112652_vessels_MMSI_type_change.js
fawadik: yeah...cool Omid..... i did check it...

src/server/seeds/dev/006_00_vessels.js
```diff
@@ -12,17 +12,17 @@ exports.seed = function(knex, Promise) {
...
         {
           organization_id: 1,
-          mmsi: 45.33,
+          mmsi: "whatever",
           name: "hyf-vessel2"
         },
         {
           organization_id: 1,
-          mmsi: 24.33,
+          mmsi: "whatever",
```
zkwsk: The seed values should numeral values. The reason I wrote it should be a string is that the numeral values are not something you want to compute - i.e. it does not make sense to do `phoneNumber + 1`, so phone numbers are usually stored as a string. There might also be situations where phone numbers might contain special characters like `+(45)`. In this case the MMSI is just 9 numbers, but it should still be stored as a string.

zkwsk (approved):
I updated the MMSI values myself, so this can be merged.

PR #83, opened 2019-09-26
loginform component

hemagunti:
Created login form component to fetch the token and storybook as well.

### DIFF
--- a/src/client/App.js
+++ b/src/client/App.js
@@ -1,12 +1,11 @@
 import React, { Component } from "react";
 import { withRouter } from "react-router-dom";
 import MainRouter from "./routes/main-router";
-import LoginForm from "./components/login-form";
+
 class App extends Component {
   render() {
     return (
       <div>
-        <LoginForm />
         <MainRouter />
       </div>
     );
--- a/src/client/App.js
+++ b/src/client/App.js
@@ -1,10 +1,15 @@
 import React, { Component } from "react";
 import { withRouter } from "react-router-dom";
 import MainRouter from "./routes/main-router";
-
+import LoginForm from "./components/login-form";
 class App extends Component {
   render() {
-    return <MainRouter />;
+    return (
+      <div>
+        <LoginForm />
+        <MainRouter />
+      </div>
+    );
   }
 }
 
--- a/src/client/components/login-form.js
+++ b/src/client/components/login-form.js
@@ -1,0 +1,75 @@
+import React, { Component } from "react";
+class LoginForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      email: "",
+      password: "",
+      result: ""
+    };
+  }
+
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
+    });
+  };
+  handleSubmit = event => {
+    event.preventDefault();
+
+    fetch("http://localhost:3000/api/auth/login/ ", {
+      method: "POST",
+      headers: {
+        Accept: "application/json text/plain",
+        "Content-Type": "application/json"
+      },
+      body: JSON.stringify({
+        email: this.state.email,
+        password: this.state.password
+      })
+    })
+      .then(response => response.json())
+      .then(data => {
+        localStorage.setItem("token", data.token); //storing token into local storage
+        console.log(data);
+      });
+  };
+  catch(err) {
+    console.log(err);
+    return err.message;
+  }
+
+  render() {
+    const { email, password, result } = this.state;
+    return (
+      <form onSubmit={this.handleSubmit}>
+        <h1>LoginForm</h1>
+        <br />
+        <label>
+          Email:
+          <input
+            name="email"
+            type="text"
+            value={email}
+            placeholder="Enter your Email"
+            onChange={this.handleChange}
+          />
+        </label>
+        <br />
+        <label>
+          Password:
+          <input
+            name="password"
+            type="password"
+            value={password}
+            onChange={this.handleChange}
+          />
+        </label>
+        <button type="submit">Login</button>
+        <button onClick={this.clickHandlerCancel}>Cancel</button>
+        <h1>{this.state.result}</h1>
+      </form>
+    );
+  }
+}
+export default LoginForm;
--- a/src/client/components/login-form.js
+++ b/src/client/components/login-form.js
@@ -8,7 +8,12 @@ class LoginForm extends React.Component {
       result: ""
     };
   }
-
+  clickHandlerCancel = event => {
+    this.setState({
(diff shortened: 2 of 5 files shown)


zkwsk (changes requested):
Not sure what's wrong, but I get this error when I try to run Storybook:

```
ERROR in ./stories/login-form.js
Module not found: Error: Can't resolve '../components/login-form/login-form' in '/Users/zaki/projects/ov-class09-fp/src/client/stories'
```

hemagunti:
Zaki, I resolved it. Can you try it now.

src/client/components/login-form.js
```diff
@@ -0,0 +1,74 @@
...
+
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
+    });
+  };
+  handleSubmit = event => {
+    event.preventDefault();
+
+    fetch("http://localhost:3000/api/auth/login/ ", {
```
bytewiz: This will only work as long as we run it locally, we should possibly keep this as a config or environment variable.
We could have a `ROOT_URL` that depends on `process.env` so we have one when running in development "mode" and another when running in production

src/client/components/login-form.js
```diff
@@ -0,0 +1,74 @@
...
+          Password:
+          <input
+            name="password"
+            type="password"
+            value={password}
+            onChange={this.handleChange}
+          />
+        </label>
+        <button type="submit">Login</button>
+        <button onClick={this.clickHandlerCancel}>Cancel</button>
```
bytewiz, line 72: this function `this.clickHandlerCancel` doesn't seem to exist

src/client/stories/login-form.js
```diff
@@ -0,0 +1,16 @@
...
+import { withKnobs, text } from "@storybook/addon-knobs";
+
+export function LoginFormStory() {
+  const email = text("Email", "test@test.com");
+  const password = text("password", "test");
+
+  return (
+    <div>
+      <LoginForm Email={email} Password={password} />
+      <Button>Login</Button>
```
bytewiz: Why is there a `Button` here? there is already a login button inside the `LoginForm`?

src/client/components/login-form.js
```diff
@@ -0,0 +1,74 @@
...
+        password: this.state.password
+      })
+    })
+      .then(response => response.json())
+      .then(data => {
+        localStorage.setItem("token", data.token); //storing token into local storage
+        //console.log(data);
+      });
+  };
+  catch(err) {
```
bytewiz: What is this catch doing? should that have a `try{}`wrapping the API call? :)

bytewiz (changes requested):
Have a few comments, also I just a thought on making this more reusable (which we don't have to) but passing in the submit function would make the login form more reusable as it wouldn't have to know about the logic of the API call :)

src/client/components/login-form.js
hemagunti: Hi Niels, Yes that's correct. Can you please provide the Development URL / Production Url(if available) so that I will try.

src/client/components/login-form.js
hemagunti: resolved it.

src/client/stories/login-form.js
hemagunti: resolved

src/client/components/login-form.js
hemagunti: I did't have clear details about cancel button actions so just created a placeholder. Now I have developed event handler for cancel button. We can change later if we need to perform new action on cancel button.

src/client/components/login-form.js
```diff
@@ -0,0 +1,78 @@
...
+  };
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
+    });
+  };
+  handleSubmit = event => {
+    event.preventDefault();
+    try {
+      fetch("http://localhost:3000/api/auth/login/ ", {
```
zkwsk: We already have that env variable!

Just use:
```
fetch(`${process.env.API_PATH}/auth/login/`, {
```

src/client/components/login-form.js
```diff
@@ -0,0 +1,78 @@
...
+        },
+        body: JSON.stringify({
+          email: this.state.email,
+          password: this.state.password
+        })
+      })
+        .then(response => response.json())
+        .then(data => {
+          localStorage.setItem("token", data.token); //storing token into local storage
+          //console.log(data);
```
zkwsk: Please remove the console.log. I know it's commented out but it's good practice to remove them entirely.

zkwsk (changes requested):

src/client/components/login-form.js
bytewiz: Yeah, I will have to find that 👍

src/client/components/login-form.js
hemagunti: removed console

src/client/components/login-form.js
hemagunti: I am getting Undefined when I am using this path. Can you pls check once

hemagunti: Hi Zaki, This path returns undefined. Can you pls check and let me know

src/client/App.js
```diff
@@ -4,7 +4,11 @@ import MainRouter from "./routes/main-router";
 
 class App extends Component {
   render() {
-    return <MainRouter />;
+    return (
+      <div>
+        <MainRouter />
+      </div>
```
Swapna-Sahu, line 10: You can remove this <div> because there is only one component.

Swapna-Sahu (commented):
Looks good other than api. I am not sure if it is correct.

src/client/App.js
zkwsk, line 10: Agree. Please remove the enclosing div.

src/client/components/login-form.js
```diff
@@ -0,0 +1,78 @@
...
+  handleChange = event => {
+    this.setState({
+      [event.target.name]: event.target.value
+    });
+  };
+  handleSubmit = event => {
+    event.preventDefault();
+
+    try {
+      fetch(`${process.env.API_PATH}/auth/login/`, {
```
zkwsk, line 26: We need to use webpack to print the env vars on compile. Like in this article: https://medium.com/@trekinbami/using-environment-variables-in-react-6b0a99d83cf5

Tried getting it to work today, but was not succesful. Will give it another shot tomorrow.

zkwsk:
Is this PR still relevant, given that you've opened? #95 that looks to contain the same changes? If not, please close so we don't merge it by accident.

PR #70, opened 2019-09-24
Api/vessel reports

Swapna-Sahu:
Renamed the column position to position_waypoints.

PR #87, opened 2019-09-26
Api/vessel reports

Swapna-Sahu:
With ref to https://github.com/HackYourFuture-CPH/ov-class09-fp/pull/70


src/server/api/routes/router.js
```diff
@@ -8,13 +8,49 @@ const userRouter = require("./user.router");
...
+router.use("/organizations", organizationRouter);
+router.use("/vessels", vesselRouter);
+// /api/organizations/organization_id/vessels
+
+// /api/voyages
+router.use("/voyages", voyagesRouter);
+
+// /api/favorite_vessels/
+router.use("/favorite-vessels", favoriteVesselsRouter);
+
```
zkwsk: You still did not remove all of the duplicate routes. 

Everything on lines :22 to :53 is the same as :66 to :77.

Only difference being that I add a variable `process.env.API_PATH` before the path. Reason is so we easily can configure the API to run on a different path. F.eks. we may want the first version of the api to run at /api/v1 and later we want to add another api running at /api/v2.

zkwsk (changes requested):

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -0,0 +1,77 @@
...
+    return `no vessel exist with that id ${id} `;
+  }
+  return vesselReports;
+};
+
+//create vessel-reports
+const createVesselReport = async ({ body }) => {
+  const {
+    vessel_id,
+    position_waypoints,
```
zkwsk: It's called `position_waypoint` in singular, not plural. It's called like that because it's a column in the DB, not a table.

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -0,0 +1,77 @@
...
+    .where({
+      id: vessel_id
+    });
+  if (vessel.length === 0) {
+    throw new HttpError("Bad request", `${vessel_id} vessel not found!`, 404);
+  }
+
+  const vesselReport = await knex("vessel_reports").insert({
+    vessel_id: vessel[0].id,
+    position_waypoints: position_waypoints,
```
zkwsk: Replace with `position_waypoint`

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+//Get vessel-reports by id
+const getVesselReportById = async id => {
+  try {
+    const vesselReport = await knex("vessel_reports")
+      .select(
+        "vessel_reports.id as id",
+        "vessel_id",
+        "position_waypoints",
```
zkwsk: Replace with `position_waypoint`

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -0,0 +1,77 @@
...
+    hfo_consumption,
+    lsfo_consumption
+  } = body;
+
+  const vessel = await knex
+    .from("vessels")
+    .select("*")
+    .where({
+      id: vessel_id
+    });
```
zkwsk, line 32: You have a problem with this query. I'm not quite sure why:
```
Error: Undefined binding(s) detected when compiling SELECT. Undefined column(s): [id] query: select `vessels`.`id` as `id` from `vessels` where `id` = ?
```

zkwsk: I spend some more time debugging this and it turns out that `body` is undefined. It's too late for me to debug further, but have a look at the router as well where it comes from.

zkwsk (changes requested):

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: replaced

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: replaced

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: I am not sure why you are not getting. But I am getting all input. Maybe be in the body of postman , you are giving "id" also. I think we should not give id in body because that has to create by itself. 

-

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: replaced

src/server/api/routes/router.js
Swapna-Sahu: removed

zkwsk (approved):
Thanks for your patience @Swapna-Sahu 👯‍♂ 👯

fawadik (commented):
looks fine for me

fawadik (approved):
it looks fine for me

Swapna-Sahu:
Merged vessel-reports

PR #89, opened 2019-09-30
vessels-details

Swapna-Sahu:
https://github.com/HackYourFuture-CPH/ov-class09-fp/pull/72

Added new components 
1.Card 
2.Card Group


bytewiz (commented):
When I run `npm start` none of the routes seem to work, does it work for you?

Nevermind: forgot it is `npm run dev`

bytewiz (dismissed):
Looks good to me! 👍

src/client/components/VesselDetails.js
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from "react";
+import CardTeaser from "./CardTeaser";
+import { CardGroup, CardGroupHeader } from "./CardGroup";
+import VesselTeaser from "./VesselTeaser";
+
+function VesselDetailsComponent({
+  vesselName,
+  departureFrom,
```
zkwsk: Would prefer `departsFrom` as it is more consistent with `arrivesAt`. Please rename all occurrences throughout the PR.

src/client/components/VesselDetails.js
```diff
@@ -0,0 +1,45 @@
...
+  arrivesAt,
+  etd,
+  eta,
+  date,
+  lat,
+  lon,
```
zkwsk: Please rename to `longitude` and `latitude` for constistency :)

zkwsk: And remember to replace throughout the PR.

src/client/components/VesselDetails.js
```diff
@@ -0,0 +1,45 @@
...
+        <CardTeaser title="DATE" data={date} />
+        <CardTeaser title="LATTITUDE" data={lat} />
+        <CardTeaser title="LONGITUDE" data={lon} />
+      </CardGroup>
+
+      <CardGroupHeader title="Consumption" />
+      <CardGroup>
+        <CardTeaser title="HSFO" data={hsfo} />
+        <CardTeaser title="ULSFO" data={ulsfo} />
+        <CardTeaser title="FOC TOTAL" data={hsfo + ulsfo} />
```
zkwsk: Please change all of these titles to "Title Case" instead of UPPERCASE. I know it is all uppercase in the mockups, but we generally use CSS to make text like that uppercase since it is only something we do for appearance.

Imagine that the same data was used to display in another app. They might not care about it being uppercase.

zkwsk: Also, I realise that since we are not doing any styling the headings will not be uppercase right now. That's intentional :)

src/client/routes/vessels-router.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+import { Route, withRouter, Link } from "react-router-dom";
+
+import VesselDetails from "./../components/VesselDetails";
+
+const VesselRouter = props => {
+  return (
+    <React.Fragment>
```
zkwsk: Nice to see you know about `Fragments`, but there is absolutely no point in using them here, since you already have a `<div>` wrapping everything.

src/client/routes/vessels-router.jsx
```diff
@@ -0,0 +1,17 @@
...
+      <div className="container">
+        <Route exact path="/vessels" component={VesselDetails} />
```
zkwsk: Should be `/` but we also discussed that on Slack.

zkwsk (changes requested):
Hey @Swapna-Sahu 👋 I'm nitpicking a little bit here, but it would be nice to have a few more things fixed before merging 🙏

src/client/components/CardTeaser.js
```diff
@@ -0,0 +1,12 @@
+import React, { Component } from "react";
+
+function CardTeaser({ title, data }) {
```
zkwsk: Please just call this one `Card` (or maybe `DataCard` - see below). It's not a "teaser" since it is not teasing anything (you cannot click it to see full details).

src/client/components/CardTeaser.js
```diff
@@ -0,0 +1,12 @@
...
+  return (
+    <div>
+      <h4>{title}</h4>
+      <p>{data}</p>
```
zkwsk: Hmmm.... could we use `props.children` to make it more flexible what you can put inside a card? 

Alternatively we might need a different name for the card to indicate that it is a "small" card with only one label and text field? `DataCard` maybe?

src/client/components/VesselDetails.js
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from "react";
+import CardTeaser from "./CardTeaser";
+import { CardGroup, CardGroupHeader } from "./CardGroup";
+import VesselTeaser from "./VesselTeaser";
+
+function VesselDetailsComponent({
```
zkwsk: Sorry - I keep finding small things 😆 Please rename to `VesselDetails`. We don't call our components "component" - we put them in the component folder instead :)

zkwsk (changes requested):

src/client/components/VesselDetails.js
Swapna-Sahu: renamed departureFrom to departsFrom

src/client/components/VesselDetails.js
Swapna-Sahu: renamed to longitude and latitude

src/client/components/VesselDetails.js
Swapna-Sahu: renamed all titles in Title Case expect HSFO,ULSFO,FOC,ETD,ETA.

src/client/routes/vessels-router.jsx
Swapna-Sahu: Removed fragment

src/client/routes/vessels-router.jsx
Swapna-Sahu: removed because not related to this task

src/client/components/CardTeaser.js
Swapna-Sahu: Renamed from card teaser to DataCard

src/client/components/VesselDetails.js
Swapna-Sahu: I am learning how to do naming convention through your comments. So good for me :)
renamed VesselDetails

src/client/components/CardTeaser.js
zkwsk: We are going with `DataCard`.

Swapna-Sahu: Decided to check again when we used Material UI.

zkwsk (approved):
Beautiful work! Thanks for hanging in there with the review 🥇 🥇 🥇

Swapna-Sahu:
Thanks Zaki for having patience.

zkwsk:
@bytewiz had previously approved, so I merged it.

PR #91, opened 2019-10-02
Refactor selected routes

NayanaKamtekar:
I have completed 
"Refactor selected routes
-Change reference suggested_routes > voyages to suggested_routes > vessel_reports
-Add column selected_routes on vessel_reports pointing at suggested_routes
-Update seed files
-Delete existing selected_routes table"
Please review.


zkwsk (approved):

zkwsk (commented):
This will break some APIs. We need to figure out what to do about that:

`api/vessels/{vessel_id}/vessel-reports`

But the logic is already broken, so I think we can still merge. And fix this in another PR.

zkwsk:
Here is the task to fix the API:

https://trello.com/c/fGRU6W1M/69-get-vessel-report-by-voyage

Will merge now, since we did a 2nd review in class.

PR #93, opened 2019-10-03
Suggested Route UI

lil2400:


src/client/components/Table.js
```diff
@@ -0,0 +1,67 @@
...
+    {
+      date: "2019-01-01 09:08:50",
+      latitude: 28.650932,
+      longitude: 86.435767,
+      speed: 19,
+      estimated_prm: 61
+    }
+  ]);
+  return (
+    <Paper>
```
zkwsk: Please remove the paper container. I'm not entirely sure about the design yet, but I'm pretty sure that there will be a parent container with either paper or card applied.

package.json
```diff
@@ -23,6 +23,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.8.1",
     "@fortawesome/react-fontawesome": "^0.1.4",
     "@hapi/joi": "^15.1.1",
+    "@material-ui/core": "^4.4.3",
```
akafaneh, line 26: is it needed to install a whole library to use one component? we will discuss using material UI tomorrow and decide handling components in general

lil2400: Thanks for taking time to review. I believe we will use material ui in all components, and here I started 😀😁

src/client/components/Table.js
JustinaOyov: Need to remove the Paper.

src/client/components/Table.js
```diff
@@ -0,0 +1,67 @@
...
+        <TableBody>
+          {data.map(row => (
+            <TableRow key={row.key}>
+              <TableCell component="th" scope="row">
+                {row.date}
+              </TableCell>
+              <TableCell align="right">{row.latitude}</TableCell>
+              <TableCell align="right">{row.longitude}</TableCell>
+              <TableCell align="right">{row.speed}</TableCell>
+              <TableCell align="right">{row.estimated_prm}</TableCell>
```
JustinaOyov: Spelling typo  {row.estimated_prm}

zkwsk (changes requested):

JustinaOyov (changes requested):
Move story in to storybook and make a separate component called SuggestedRoutesTable.

PR #94, opened 2019-10-04
Display route alternatives

ghost:


src/client/components/SuggestedRoutes/SuggestedRouteCard.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import SuggestedRouteDetails from "./SuggestedRouteDetails";
+import SuggestedRouteSummary from "./SuggestedRouteSummary";
+
+export default function SuggestedRouteCard(props) {
```
akafaneh: why not destructuring  the props here to make it shorter and more readable line?

src/client/components/SuggestedRoutes/SuggestedRouteDetails.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from "react";
+import TitleAndDescribe from "../TitleAndDescribe";
+
+export default class SuggestedRouteDetails extends Component {
```
akafaneh: why this component is class based? why not a functional component? it doesn't  hold any state of it's own or need life cycle hook

src/client/components/SuggestedRoutes/SuggestedRouteSummary.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from "react";
+import TitleAndDescribe from "../TitleAndDescribe";
+
+export default class SuggestedRouteSummary extends Component {
```
akafaneh: again this one, why is a class component? a rule of thumb: "don't use it if you don't need it". class based components are way heavier than functional one and if you are not using state or life cycle hooks then you don't need it. (out of scope but good to know, with the new react version we don't use classes no more as react introduced [hooks](https://reactjs.org/docs/hooks-intro.html) to have state and side effects on functional components.)

src/client/components/SuggestedRoutes/SuggestedRouteSummary.js
```diff
@@ -0,0 +1,18 @@
...
+  render() {
+    const { id, isSelected, handleClick, time, price } = this.props;
+    return (
+      <div>
+        {isSelected ? <h4>SELECTED ROUTE</h4> : null}
+        <TitleAndDescribe title="HOURS" description={time} />
+        <TitleAndDescribe title="TOTAL COST" description={price} />
+        {!isSelected ? (
+          <button onClick={() => handleClick(id)}>SELECT ROUTE</button>
```
akafaneh: this is not wrong, but a bad practice/habit, always try to strip out functions out of the render scope if you can.
```
const onSelectRoutClick = () => {
    this.props.handleClick(this.props.id)
}
  render() {
    const { id, isSelected, handleClick, time, price } = this.props;
    return (
      <div>
        {isSelected ? <h4>SELECTED ROUTE</h4> : null}
        <TitleAndDescribe title="HOURS" description={time} />
        <TitleAndDescribe title="TOTAL COST" description={price} />
        {!isSelected ? (
          <button onClick={onSelectRoutClick}>SELECT ROUTE</button>
```
the issue with the way you did it, is that on every render you create a new whole function, when you strip the arrow function out like above, it get created once on the fist load and it just keep referring to it on each render after that, which is good for performance.

src/client/components/SuggestedRoutes/SuggestedRoutesList.js
```diff
@@ -0,0 +1,40 @@
...
+import SuggestedRouteCard from "./SuggestedRouteCard";
+
+export default class SuggestedRoutesList extends Component {
+  state = {
+    selectedId: null
+  };
+
+  handleClick = id => {
+    this.setState({
+      ...this.state,
```
akafaneh: why you are spreading the old state here before assigning a new one?

src/client/components/SuggestedRoutes/SuggestedRoutesList.js
```diff
@@ -0,0 +1,40 @@
...
+      selectedId: id
+    });
+  };
+
+  render() {
+    return (
+      <div>
+        {this.props.list.map(i => {
```
akafaneh: it's alway better to use more meaningful names, what is a list here? what is i refers to? it good naming make the code more understandable

src/client/components/SuggestedRoutes/SuggestedRouteSummary.js
akafaneh: this is not wrong, but a bad practice/habit, always try to strip out functions out of the render scope if you can.
```
onSelectRoutClick = () => {
    this.props.handleClick(this.props.id)
}
  render() {
    const { id, isSelected, handleClick, time, price } = this.props;
    return (
      <div>
        {isSelected ? <h4>SELECTED ROUTE</h4> : null}
        <TitleAndDescribe title="HOURS" description={time} />
        <TitleAndDescribe title="TOTAL COST" description={price} />
        {!isSelected ? (
          <button onClick={this.onSelectRoutClick}>SELECT ROUTE</button>
```
the issue with the way you did it, is that on every render you create a new whole function, when you strip the arrow function out like above, it get created once on the fist load and it just keep referring to it on each render after that, which is good for performance.

src/client/components/SuggestedRoutes/SuggestedRouteDetails.js
akafaneh: why this component is class based? why not a functional component? it doesn't hold any state of it's own or need life cycle hook

src/client/components/SuggestedRoutes/SuggestedRouteCard.js
akafaneh: why not destructuring the props here to make it shorter and more readable line?

akafaneh (changes requested):
nice work over all, i left some comments you can check.

src/client/components/SuggestedRoutes/SuggestedRoutesList.js
zkwsk: I always recommend doing that. You have to pass a full copy of the state right? To preserve immutability?

I know in this case we only have `selectedId` so it won't make a difference, but in case more variables are added to the state later.

src/client/components/SuggestedRoutes/SuggestedRoutesList.js
zkwsk: You mean "what is a list item here", right? Because I agree that `i` is not an idea variable name.

src/client/components/SuggestedRoutes/SuggestedRoutesList.js
zkwsk: Looks like I picked up this habit from using Redux. You should safely be able to remove line 10.

akafaneh: but setState does this out of the box :D

src/client/components/SuggestedRoutes/SuggestedRouteDetails.js
```diff
@@ -0,0 +1,30 @@
...
+      eta,
+      sailDays,
+      maxWaveHs,
+      avgSpeed,
+      hireCost,
+      consumption
+    } = this.props;
+
+    return (
+      <React.Fragment>
```
zkwsk: Tip: for fragments if you write your import like this:
```
import React, { Fragment } from "react";
```
You will be able to write:
```
<Fragment>...</Fragment>
```

Personally I think that looks a lot cleaner. There is also the shorthand `<>...</>` syntax, but personally I think that looks like a typo so that hurts my eyes a little :D But this is just a matter of personal preference, so no need to change it :)

src/client/components/TitleAndDescribe.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+export default function TitleAndDescribe({ title, description }) {
```
zkwsk: As we discussed on Slack, I really don't like the name `TitleAndDescribe`. What if I choose that all of them should have a unit (meter, dollars, etc.), should it then be called `TitleAndDescribeAndUnits`.

So think about the purpose of the component. It is essentially a component that shows a very small chunk of information with a describing label. The best name I could come up with was `Field`. It is very closely related to the component `DataCard` that @Swapna-Sahu did though. I think at some point the `DataCard` should just be a regular `Card` component wrapping this component. But I need to look a little more into Material UI to decide this finally. For now, I will be happy if you call this one `Field`.

src/client/stories/suggestedRoutes/suggestedRouteCard.js
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <SuggestedRouteCard
+      time={text("theTime")}
+      price={text("thePrice")}
+      eta={text("ETA")}
+      sailDays={text("SAIL DAYS")}
+      maxWaveHs={text("MAX WAVE HS")}
+      avgSpeed={text("AVG SPEED")}
+      hireCost={text("HIRE COST")}
+      consumption={text("CONSUMPTION")}
```
zkwsk: A small thing:

These are the labels you see in Storybook. It's not that important, but to satisfy my OCD it would be nice if you will make the labels a little more consistent:
```suggestion
time={text("Time")}
price={text("Price")}
eta={text("ETA")}
sailDays={text("Sail days")}
maxWaveHs={text("Max wave HS")}
avgSpeed={text("Average speed")}
hireCost={text("Hire cost")}
consumption={text("Consumption")}
```

zkwsk: Note that it is only abbreviations that are capitalized.

src/client/stories/suggestedRoutes/suggestedRouteDetails.js
```diff
@@ -0,0 +1,18 @@
...
+
+export function suggestedRouteDetails() {
+  return (
+    <SuggestedRouteDetails
+      eta={text("ETA")}
+      sailDays={text("SAIL DAYS")}
+      maxWaveHs={text("MAX WAVE HS")}
+      avgSpeed={text("AVG SPEED")}
+      hireCost={text("HIRE COST")}
+      consumption={text("CONSUMPTION")}
```
zkwsk: Again, consistency can be improved. See above.

zkwsk (changes requested):

src/client/stories/suggestedRoutes/suggestedRouteCard.js
ghost: @zkwsk I didn't understand what should I do exactly!

src/client/components/SuggestedRoutes/SuggestedRoutesList.js
ghost: @zkwsk what is the final decision!

src/client/stories/suggestedRoutes/suggestedRouteDetails.js
ghost: I didn't understand here as well!

ghost:
@akafaneh Thanks a lot to review my code, I tried my best to fix the issues.

src/client/components/SuggestedRoutes/SuggestedRouteSummary.js
ghost: I tried what you mentioned here but I got some errors and I think if it not neccecry I would like to leave it like this otherwise I need your help.

src/client/components/SuggestedRoutes/SuggestedRoutesList.js
zkwsk: It's line 11 you can remove.

src/client/stories/suggestedRoutes/suggestedRouteDetails.js
ghost: Done ;)

src/client/stories/suggestedRoutes/suggestedRouteCard.js
ghost: Done ;)

src/client/components/SuggestedRoutes/SuggestedRouteSummary.js
ghost: Done ;)

src/client/components/Field.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+export default function Field({ title, description }) {
+  return (
+    <div>
+      <h4>{title}</h4>
+      <h4>{description}</h4>
```
Swapna-Sahu: Are your sure title and description both should be h4

src/client/stories/suggestedRoutes/suggestedRouteCard.js
```diff
@@ -0,0 +1,27 @@
...
+  number,
+  array,
+  object
+} from "@storybook/addon-knobs";
+
+export function suggestedRouteCard() {
+  return (
+    <SuggestedRouteCard
+      time={text("Time")}
+      price={text("Price")}
```
Swapna-Sahu: Price is a text or number?

src/client/stories/suggestedRoutes/suggestedRouteCard.js
```diff
@@ -0,0 +1,27 @@
...
+      eta={text("ETA")}
+      sailDays={text("Sail days")}
```
Swapna-Sahu: Sails days also I think in number

src/client/stories/suggestedRoutes/suggestedRouteCard.js
```diff
@@ -0,0 +1,27 @@
...
+      maxWaveHs={text("Max wave HS")}
+      avgSpeed={text("Average speed")}
+      hireCost={text("Hire cost")}
+      consumption={text("Consumption")}
```
Swapna-Sahu: please check all prop type for all

Swapna-Sahu (changes requested):
I have few comments. Please check .

src/client/components/Field.js
ghost: @Swapna-Sahu 
No, but we can style them later, what is your suggestion?

src/client/stories/suggestedRoutes/suggestedRouteCard.js
ghost: This is going to be just in Storybook to show the demo, So I don't think it is that much important to change it to number.
but if @zkwsk asks me I can change it!

@Swapna-Sahu thanks for mentioning.

src/client/stories/suggestedRoutes/suggestedRouteCard.js
ghost: yes, but it will be in storyBook to show the demo!

src/client/stories/suggestedRoutes/suggestedRouteCard.js
ghost: @zkwsk Please let me know if it in necessary to change them?

src/client/stories/suggestedRoutes/suggestedRouteCard.js
zkwsk: True, but when we add propTypes checking this will fail in Storybook. So please change it to number.

src/client/components/Field.js
zkwsk: I would suggest:
```suggestion
      <h3>title</h3>
      <p>{description}<p>
```

zkwsk (changes requested):
Hey @Omid-Incarnation!

Since we just added prop-type validation, please set up prop-types for your components before we merge them.

Check Abods PR how to do it here:
https://github.com/HackYourFuture-CPH/ov-class09-fp/pull/101

https://github.com/HackYourFuture-CPH/ov-class09-fp/pull/101

src/client/components/Field.js
ghost: Done :)

src/client/stories/suggestedRoutes/suggestedRouteCard.js
```diff
@@ -0,0 +1,20 @@
...
+    <SuggestedRouteCard
+      time={number("Time", "26h / +6")}
+      price={number("Price", "$104.207/ -$6.208")}
+      eta={text("ETA", "2020-03-02 06:50:33")}
+      sailDays={number("Sail days", "6.5")}
+      maxWaveHs={number("Max wave HS", "11.52")}
+      avgSpeed={number("Average speed", "23.52")}
+      hireCost={number("Hire cost", "$66.808")}
+      consumption={number("Consumption", "$60.600")}
+    />
```
zkwsk, line 21: Great work with the proptypes.

Now you just need to make sure that your stories are passing in all of the prop-types you declared. When I open storybook and open the console I see a lot of errors because here you are for example not passing in an ID or a `isSelected`prop.

Please go over this story as well as the other stories you've added and make sure all props are being fed in correctly :)

zkwsk (changes requested):

src/client/stories/suggestedRoutes/suggestedRouteCard.js
zkwsk: That probably also means removing the units from the numbers above as we discussed, because you can't pass "$66.808" into a number knob for example.

zkwsk (changes requested):

zkwsk (dismissed):

Swapna-Sahu (dismissed):
Looks good to me but need to exchange the suggested route summary and suggested route card.

PR #95, opened 2019-10-04
Protected routes

hemagunti:
To check if the user is logged in in order to view certain routes and if they do not redirect them to the login page.

and also if they did login before that we check they are already logged in and then we redirect them into the app.


src/client/components/AppLayout.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { Route, Redirect } from "react-router-dom";
+class AppLayout extends React.Component {
```
akafaneh: this component doesn't include state or life cycle hook, so maybe using functional component is lighter, don't you think so?

src/client/components/PrivateRoute.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Route, Redirect } from "react-router-dom";
+
+const PrivateRoute = ({ component: Component, ...rest }) => {
+  const isLoggedin = localStorage.getItem("token");
```
akafaneh, line 5: aren't you supposed to check of it's a valid token to assure that the user is actually logged in?

src/client/components/login-form.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  clickHandlerCancel = event => {
+    this.setState({
+      email: "",
+      password: ""
+    });
+  };
+
+  handleChange = event => {
```
akafaneh, line 19: i like this generic change handler! great job

src/client/components/login-form.js
```diff
@@ -0,0 +1,100 @@
...
+      [event.target.name]: event.target.value
+    });
+  };
+
+  handleSubmit = event => {
+    event.preventDefault();
+
+    try {
+      // TODO: Add environment variable
+      fetch("api/auth/login", {
```
akafaneh: maybe services/api-calls should be in a separate utility-like js file? we can talk about this tomorrow. maybe also you should use [axios](https://github.com/axios/axios)?

src/client/components/login-form.js
```diff
@@ -0,0 +1,100 @@
...
+
+  render() {
+    const { email, password, isLoggedin } = this.state;
+    const { from } = this.props.location.state || {
+      from: { pathname: "/home" }
+    };
+
+    if (isLoggedin) {
+      return <Redirect to={from} />;
+    } else {
```
akafaneh: no need for this else as return breaks out from the function

src/client/components/login-form.js
```diff
@@ -0,0 +1,100 @@
...
+      return (
+        <div>
+          <p>You must log in to view this page at {from.pathname}</p>
+          <form onSubmit={this.handleSubmit}>
+            <h1>LoginForm</h1>
+            <br />
+            <label>
+              Email:
+              <input
```
akafaneh: maybe we should consider making a reusable input components (or just introduce the idea of reusable components?

akafaneh (commented):
Nice work over all, check the comments and we can talk about them and see how we can improve the code

src/client/components/PrivateRoute.js
zkwsk: We don't actually have an endpoint for this at the moment. But all of the API endpoints except for the one to login are protected, so we discussed barely checking the existence of a token would at least stop the user from going into the application without having access to any content.

src/client/components/login-form.js
```diff
@@ -0,0 +1,100 @@
...
+import { Route, Redirect } from "react-router-dom";
+
+class LoginForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      email: "",
+      password: "",
+      isLoggedin: false
+    };
```
zkwsk: I really like using class fields for this:
```suggestion
state = {
  ...
}
```
No need to make a constructor, then.

src/client/components/login-form.js
zkwsk: Might be a good idea. `fetch()` some times acts a little strange with error-handling. I just feel like `fetch()` is good to learn as it will probably be the way of the future, as it is a native browser api. But I agree that we would probably replace it with axios eventually.

src/client/components/login-form.js
zkwsk: Yeah. I think we might have some of the input fields already from the start voyage form.

src/client/components/login-form.js
```diff
@@ -0,0 +1,100 @@
...
+              Email:
+              <input
+                name="email"
+                type="text"
+                value={email}
+                placeholder="Enter your Email"
+                onChange={this.handleChange}
+              />
+            </label>
+            <br />
```
zkwsk: We don't want any `<br>` tags. We will make it pretty later.

src/client/routes/main-router.jsx
```diff
@@ -1,14 +1,19 @@
...
+import LoginForm from "./../components/login-form";
+import PrivateRoute from "./../components/PrivateRoute";
+import AppLayout from "./../components/AppLayout";
 
 const MainRouter = props => {
   return (
     <React.Fragment>
       <div className="container">
-        <Route exact path="/" component={Home} />
+        <Route exact path="/home" component={Home} />
```
zkwsk: The path is `/` not `/home`. The component however is called `Home`.

zkwsk (changes requested):
I could not quite make it out in your code, but if you try to login and enter a wrong password we are still redirected to `/home`. We should stay on `/login`. This is really important, because otherwise we are not really checking if the user is logged in.

src/client/components/AppLayout.js
hemagunti: yes true

src/client/components/login-form.js
hemagunti: removed constructor.

src/client/components/login-form.js
hemagunti: removed <br> tag

src/client/routes/main-router.jsx
hemagunti: resolved the comment.

src/client/components/login-form.js
hemagunti: comment resolved

src/client/components/login-form.js
hemagunti: I have replaced axis instead of fetch

zkwsk (approved):

PR #96, opened 2019-10-04
ger Users by organization Id

ghost:
By the API below 
`//ENDPOINT /api/organizations/:organization_id/users/ :get`

you can have a list of users in the same Organization!


src/server/api/controllers/user.controller.js
```diff
@@ -321,6 +321,16 @@ const createAdminBySuperUser = async (id, body) => {
   });
 };
 
+const gerUsersbyorganizationId = async organization_id => {
```
Swapna-Sahu: Can you rename this "gerUsersbyorganizationId " to "getUsersByOrganizationId"

Swapna-Sahu (changes requested):
Only one small change needed. Other than that all looks good to me.

src/server/api/controllers/user.controller.js
ghost: @Swapna-Sahu thanks for mentioning this...

src/server/api/controllers/user.controller.js
```diff
@@ -321,6 +321,16 @@ const createAdminBySuperUser = async (id, body) => {
   });
 };
 
+const getUsersbyOrganizationId = async organization_id => {
+  const usersByOrganizationId = await knex
+    .from("users")
+    .where({ organization_id });
+  if (usersByOrganizationId.length === 0) {
+    return `This Organization have no users!`;
```
ismailirahim: Throw an HTTP error here instead of just returning a string! :heart:

ismailirahim (changes requested):
changes reuqest! :boom:

src/server/api/controllers/user.controller.js
ghost: Done ;)

AbodHassam (dismissed):
Looks good to me

zkwsk (approved):
I pushed a small fix to swagger myself, but other than that it looks good 👍

PR #97, opened 2019-10-05
Update db diagram

AbodHassam:


db/Schema.dbml
```diff
@@ -1,57 +1,65 @@
...
+  is_locked int [ default: null]
+}
+
+table knex_migrations {
+  id int [ pk, not null, increment ]
+  name varchar [ default: null]
+  batch int [ default: null ]
+  migration_time timestamp [ default: null]
+}
+
```
zkwsk: Sorry @AbodHassam, I did not think about this, but we do not want the Knex tables in our diagram. That's not really relevant to our app and just adding noise :)

db/Schema.dbml
```diff
@@ -155,31 +147,29 @@ enum voyage_status {
...
 // A voyage has 1 arrival waypoint
 ref: voyages.arrive_at > waypoints.id
 // A waypoint may belong to 0-* vessel reports
-ref: waypoints.id < vessel_reports.position 
-// Selected route represents a many-to-many join table between route and voyage
-// Current selected route is found by looking for the latest record
-ref: suggested_routes.id < selected_routes.fk_route_id
-ref: voyages.id < selected_routes.fk_voyage_id
-ref: favorite_vessels.user_id > users.id
-ref: favorite_vessels.vessel_id > vessels.id
```
zkwsk, line 185: You removed the references between favorite_vessels, users and vessels? Is that because we don't have them refenced in the DB or is it an accident?

zkwsk (changes requested):
Just a couple of small changes 👍

zkwsk (changes requested):
Looks good with the dbdiagram and the dbml file, but remember you also need to check in a new screenshot. Sorry that this process is such a hassle :(

zkwsk (approved):

zkwsk:
Need this updated so I'm going to merge with 1 review.

PR #100, opened 2019-10-08
Perform a check to see if you are commmitting to a protected branch.

zkwsk:
If you try to commit something on either the `develop` or `master` branch you will now get a little warning message and the opportunity to cancel before doing so.

Does not work with merging stuff into protected branches. I see there is a `post-merge` hook, but that's too late. Maybe someone else will look into if that can be done.

Video demo!

https://www.dropbox.com/s/ebkp7kz3ec5ajke/protect-loal-branch.flv?dl=0

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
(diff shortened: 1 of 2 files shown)


AbodHassam (approved):
Cool

PR #101, opened 2019-10-08
Adding prop-types

AbodHassam:


src/client/components/DataCard.js
```diff
@@ -9,4 +10,9 @@ function DataCard({ title, data }) {
   );
 }
 
+DataCard.propTypes = {
+  title: PropTypes.string,
+  data: PropTypes.string
```
Swapna-Sahu: I am not sure if we are using this only for string. It can be number or decimal. Please check with Zaki.

src/client/components/VesselsList.js
```diff
@@ -18,4 +19,14 @@ function VesselsList({ vessels = [] }) {
   return <div>{vesselsList}</div>;
 }
 
+VesselsList.propTypes = {
+  vessels: PropTypes.array,
+  id: PropTypes.string,
```
Swapna-Sahu: Id will be number not string

src/client/components/VesselTeaser.js
```diff
@@ -14,4 +15,12 @@ function VesselTeaser({ id, vesselName, departureFrom, arrivesAt, etd, eta }) {
...
 }
 
+VesselTeaser.propTypes = {
+  vesselName: PropTypes.string,
+  departureFrom: PropTypes.string,
+  etd: PropTypes.string,
+  arrivesAt: PropTypes.string,
+  eta: PropTypes.string
+};
+
```
Swapna-Sahu: I think you missed "id " here.

src/client/components/CostWidget.js
```diff
@@ -16,4 +17,9 @@ function CostWidget({ currentCost, remainingCost }) {
   );
 }
 
+CostWidget.propTypes = {
+  currentCost: PropTypes.string,
+  remainingCost: PropTypes.string
```
Swapna-Sahu: I dont think currentCost and remainingCost is a string. It should be number. Please check with Zaki.

Swapna-Sahu (changes requested):
Needed few changes and also few changes I am not sure but you can check with @zkwsk

src/client/components/CardGroup.js
```diff
@@ -1,11 +1,20 @@
 import React, { Component } from "react";
+import PropTypes from "prop-types";
 
 function CardGroupHeader({ title }) {
   return <h3>{title}</h3>;
 }
 
+CardGroupHeader.propTypes = {
+  title: PropTypes.string
```
zkwsk: Please make this prop required.

src/client/components/CardGroup.js
```diff
@@ -1,11 +1,20 @@
...
+};
+
 function CardGroup({ children }) {
   return <div>{children}</div>;
 }
 
+CardGroup.propTypes = {
+  children: PropTypes.array
```
zkwsk: Please make this prop required. Additionally you can use `propTypes.shape` to check the objects inside the array. It would be nice to make sure that you are passing `DataCard`s into the `CardGroup`.

src/client/components/CostWidget.js
zkwsk: Correct. Any cost fields should be a number.

src/client/components/VesselDetails.js
```diff
@@ -42,4 +43,16 @@ function VesselDetails({
...
 }
 
+VesselDetails.propTypes = {
+  vesselName: PropTypes.string,
+  departsFrom: PropTypes.string,
+  arrivesAt: PropTypes.string,
+  etd: PropTypes.string,
+  eta: PropTypes.string,
+  latitude: PropTypes.string,
+  longitude: PropTypes.string,
```
zkwsk: latitude/longitude are numbers

src/client/components/CostWidget.js
zkwsk: We will probably add the ability to display currency, but that should be a separate prop instead of making the number into a string.

src/client/components/DataCard.js
zkwsk: Actually it should be a string. But that means that `CostWidget` needs to pass it down as a string not as a number, so we might need to change that component. Or add different classes of DataCards. Let's make it a string for now and see where it takes us.

src/client/components/Map.js
```diff
@@ -27,4 +28,18 @@ function MapComponent({
...
+MapComponent.propTypes = {
+  centerMapCoordinates: PropTypes.array,
+  pathColor: PropTypes.string,
+  displayMarkers: PropTypes.bool,
+  zoom: PropTypes.number,
+  waypoints: PropTypes.arrayOf(
+    PropTypes.shape({
+      lat: PropTypes.number,
+      lng: PropTypes.number,
+      message: PropTypes.string
```
zkwsk: Hmm... this is Niels sample component and in storybook he is actually passing in something called `waypointMessage` but he is destructuring message here. In any case it is not working and we don't plan to have messages on the waypoints themselves (I deliberately try to keep the map as simple as possible), so could you please just remove the message propType, remove where it is being destructured off the waypoint element (:22) and remove the data from the story in storybook?

src/client/components/Marker.js
```diff
@@ -12,4 +13,10 @@ const MapMarker = ({ position, children }) => (
     {children}
   </Marker>
 );
+
+MapMarker.propTypes = {
+  position: PropTypes.array,
+  children: PropTypes.object
```
zkwsk, line 19: Here I would normally ask you to test a level deeper, but I think we are going to change this component quite a bit soon, so stick with what we have here 👍

src/client/components/VesselDetails.js
```diff
@@ -42,4 +43,16 @@ function VesselDetails({
   );
 }
 
+VesselDetails.propTypes = {
+  vesselName: PropTypes.string,
+  departsFrom: PropTypes.string,
+  arrivesAt: PropTypes.string,
+  etd: PropTypes.string,
```
zkwsk: Soon to be changed to `departure_time`. Keep a look out if this ends up being merged before or after Nayanas database refactor.

src/client/components/VesselTeaser.js
```diff
@@ -14,4 +15,12 @@ function VesselTeaser({ id, vesselName, departureFrom, arrivesAt, etd, eta }) {
   );
 }
 
+VesselTeaser.propTypes = {
+  vesselName: PropTypes.string,
+  departureFrom: PropTypes.string,
+  etd: PropTypes.string,
```
zkwsk: Again, like above, this one will soon be changed to `departure_time`.

src/client/components/VesselsList.js
```diff
@@ -18,4 +19,14 @@ function VesselsList({ vessels = [] }) {
   return <div>{vesselsList}</div>;
 }
 
+VesselsList.propTypes = {
+  vessels: PropTypes.array,
```
zkwsk: Would be nice to test if it is `Vessel` components we have in the array.

src/client/components/VesselsList.js
```diff
@@ -18,4 +19,14 @@ function VesselsList({ vessels = [] }) {
...
+  id: PropTypes.string,
+  vesselName: PropTypes.string,
+  departureFrom: PropTypes.string,
+  arrivesAt: PropTypes.string,
+  etd: PropTypes.string,
```
zkwsk: Once again, this one will soon be changed to departure_time.

zkwsk (changes requested):

src/client/components/CardGroup.js
AbodHassam: That means that i have to make DataCard props also required, right ?

src/client/components/VesselDetails.js
AbodHassam: Ok

zkwsk (approved):
Nice work 🥇 I merged develop and fixed some conflicts for you.

AbodHassam:
Thanks

PR #104, opened 2019-10-09
Api Endpoint-Get user by Role created

fawadik:
Api endpoint created for getting users by role

### DIFF
--- a/src/server/api/controllers/user.controller.js
+++ b/src/server/api/controllers/user.controller.js
@@ -33,6 +33,21 @@ const getUserById = async id => {
   }
 };
 
+const getUsersByRole = async role => {
+  try {
+    const users = await knex("users")
+      .join("user_roles", "user_roles.id", "users.role_id")
+      .select("*")
+      .where({ role: role });
+    if (users.length === 0) {
+      throw new HttpError("Bad request", `Cannot find user for ${role}!`, 404);
+    }
+    return users;
+  } catch (err) {
+    return err.message;
+  }
+};
+
 const getAccount = async req => {
   const { id } = req.user;
   try {
@@ -331,6 +346,7 @@ module.exports = {
   changePasswordRandomly,
   deleteUser,
   getUserById,
+  getUsersByRole,
   editUserRole,
   createAdminBySuperUser
 };
--- a/src/server/api/routes/user-roles.router.js
+++ b/src/server/api/routes/user-roles.router.js
@@ -1,0 +1,28 @@
+"use strict";
+
+// router setup
+
+const express = require("express");
+const router = express.Router({
+  mergeParams: true
+});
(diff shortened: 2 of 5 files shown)


src/server/api/routes/user.router.js
```diff
@@ -23,6 +23,18 @@ router.get("/", authorizeUser(ROLES.SUPER_USER), (req, res, next) => {
     .catch(next);
 });
 
+// ENDPOINT: /api/users/:role :GET
```
ismailirahim: Fowad, i was talking with @zkwsk about the endpoint and that should be /api/user-roles/:user_role_id/users

ismailirahim (changes requested):
change the endpoint to /api/user-roles/:user_role_id/users

src/server/api/routes/user.router.js
fawadik: rahim... user_role_id or only role?
bcoz zaki said to me that instead of id it should be on role

zkwsk: I'm sorry if I said something wrong. I see that it was also not described in the Trello at all. But Rahim is right about the path.

fawadik: I have created the new router and now modified the right path.
please have a look.

Swapna-Sahu (dismissed):
Looks good to me

src/server/api/controllers/user.controller.js
```diff
@@ -33,6 +33,21 @@ const getUserById = async id => {
   }
 };
 
+const getUsersByRole = async role => {
+  try {
+    const users = await knex("users")
+      .join("user_roles", "user_roles.id", "users.role_id")
+      .select("*")
+      .where({ role: role });
```
zkwsk, line 41: You need to query by the `role_id`. Currently it does not work.

```suggestion
      .where({ role_id: role });
```

src/server/api/routes/user-roles.router.js
```diff
@@ -0,0 +1,28 @@
...
+const { authorizeUser } = require("../lib/middleware/auth.middleware");
+
+const ROLES = require("../../constants/roles");
+
+// controllers
+const userController = require("../controllers/user.controller");
+
+// ENDPOINT: /api/user-roles/:role/users :GET
+router.get(
+  "/:role/users",
```
zkwsk, line 18: Can you please add documentation for this endpoint to Swagger?

zkwsk (changes requested):

src/server/api/controllers/user.controller.js
zkwsk: Sorry I was using the wrong request body. This works with role name (i.e. "admin")

zkwsk (approved):

PR #106, opened 2019-10-09
Created UI for showing All Voyages

lil2400:
Fallow the orginal ui https://trello.com/c/SX2hGt8b/54-all-voyages-page


zkwsk (changes requested):
I would like to see this broken down more. The ongoing and completed and completed table are showing almost the same, so could we break that into a `VoyagesTable` component. Then create another component wrapping this component called `AllVoyages`. The `VoyagesTable` will be a "dumb" component without state and the `AllVoyages` component will be a stateful component. To mock the stateful component, look at what Rahim is doing with the Admin interface, mocking api calls.

Swapna-Sahu (approved):
looks good to me

zkwsk (approved):

PR #110, opened 2019-10-10
Fixed waypoint methods related to new DB changes 

ismailirahim:


src/server/api/controllers/waypoints.controller.js
```diff
@@ -23,7 +23,7 @@ const getWaypointsById = async id => {
 // Method for creating a waypoints
 
 const createWaypoints = async ({ body }) => {
-  const { latitude, longitude, suggested_route_id } = body;
+  const { latitude, longitude, suggested_route_id, speed, rpm } = body;
```
Swapna-Sahu, line 26: Sequence_id is missing

src/server/api/controllers/waypoints.controller.js
```diff
@@ -43,6 +43,8 @@ const createWaypoints = async ({ body }) => {
   return await knex("waypoints").insert({
     latitude,
     longitude,
+    speed,
+    rpm,
     suggested_route_id: routes[0].id
```
Swapna-Sahu, line 48: Sequence_id is missing

Swapna-Sahu (changes requested):
Please check sequence_id before merging. Other than this all looks good for me.

src/server/api/controllers/waypoints.controller.js
zkwsk: Actually not. Since this method is for creating a single "standalone" waypoint, not a set of waypoints as part of a route, it will not make sense to declare a sequence_id.

When we are creating a suggested_route, however, we should be able to add a series of waypoints and set their sequence ID.

src/server/api/controllers/waypoints.controller.js
zkwsk: see above comment

zkwsk (approved):

Swapna-Sahu (approved):

PR #113, opened 2019-10-11
Removing pre-commit hook

zkwsk:


zkwsk:
Turns out that the pre-commit script breaks on PC, so I'm rushing this one through.

PR #114, opened 2019-10-11
Voyages by organization2

JustinaOyov:
Added Endpoint:  "Get Voyages by Organization", 
 - created a new route & returned list of voyages based on the organization and the voyage status. 
 - updated Swagger.json to add the changes. 
 - changes tested in both Swagger & Postman.


src/server/api/controllers/voyages.controller.js
```diff
@@ -53,8 +53,67 @@ const getVoyagesByVesselId = async vessel_id => {
...
+    const voyages = await knex("organizations as org")
+      .join("vessels as ves", "ves.organization_id", "org.id")
+      .where("ves.organization_id", organizationID)
+      .join("voyages as voy", "voy.vessel_id", "ves.id")
+      .where("voy.status", voyageStatus)
+      .join("ports as p1", "p1.id", "voy.depart_from_port")
+      .join("ports as p2", "p2.id", "voy.arrive_at_port")
+      .select(
+        "voy.id as id",
+        "ves.id as vessel",
```
zkwsk:
```suggestion
        "ves.id as vessel_id",
```

src/server/api/controllers/voyages.controller.js
```diff
@@ -53,8 +53,67 @@ const getVoyagesByVesselId = async vessel_id => {
...
+      );
+
+    if (voyages.length === 0) {
+      throw new HttpError(
+        "Bad request",
+        `There are no ${voyageStatus} Voyages for Organization with ID ${organizationID}!`,
+        404
+      );
+    }
+    return { voyages: voyages };
```
zkwsk: Would be equivalent to:

```suggestion
    return { voyages };
```

zkwsk (changes requested):
Great work!

Only some very small things to change. I did however notice that we need a little more functionality in this endpoint, but let's get this merged first and I'll add another story for that.

src/server/api/controllers/voyages.controller.js
JustinaOyov: Comment resolved, vessel changed to vessel_id

src/server/api/controllers/voyages.controller.js
JustinaOyov: Change incorporated.

zkwsk (approved):

Swapna-Sahu (approved):
Looks good to me

PR #115, opened 2019-10-11
Update vessels has been fixed

AbodHassam:


src/server/api/controllers/vessel.controller.js
```diff
@@ -68,14 +68,12 @@ const editVessel = async ({ body }) => {
   const vessel = await knex
     .from("vessels")
     .select("*")
```
ghost: You can delete `.select(*)` as well, because in knex you don't need it.

You can take a look at this link: [Knex cheatsheet](https://devhints.io/knex)

AbodHassam: Ok ... Thanks

Swapna-Sahu (approved):
Looks good to me

zkwsk (approved):

PR #118, opened 2019-10-12
Fix create suggested routes

zkwsk:
Sorry. Couldn't stop myself and ended up doing this one as I was debugging something else.

Updated the endpoint to reflect that suggested_routes are now associated with voyages and wrapped everything in a transaction as we should not create suggested_routes if creating voyages fail.


ismailirahim (approved):
Looks good to me

PR #123, opened 2019-10-14
vesselReports-voyage

Swapna-Sahu:
dont merge


Swapna-Sahu:
EX- /api/voyages/:voyage_id/vessel_reports?limit=1 GET

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -3,13 +3,84 @@
...
+
+    if (vesselReports.length === 0) {
+      throw new HttpError(
+        "Bad request",
+        `No vessel-reports exist with that voyage id ${voyage_id}.`,
+        404
+      );
+    }
+    return vesselReports;
+  }
```
zkwsk: You are making it a lot more complicated than it needs to be. 

I would just set default value on all of my query parameters and then do _a single_ knex statement that works no matter what the params are instead of all of the if-else spaghetti 😄 

```
const getVesselsReportByVoyageIdWithParams = async (voyage_id, query) => {
  const { limit = 50, offset = 0, orderBy = "descending" } = query;

  ...

  const vesselReports = await knex
    .from("vessel_reports")
    .where({ voyage_id })
    .orderBy("id", { orderBy })
    .limit({ limit })
    .offset({ offset })
}
```

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -3,13 +3,84 @@
...
+      .offset(0);
+
+    if (vesselReports.length === 0) {
+      throw new HttpError(
+        "Bad request",
+        `No vessel-reports exist with that voyage id ${voyage_id}.`,
+        404
+      );
+    }
+    return vesselReports;
```
zkwsk: All of this code is unreachable.

zkwsk (changes requested):

src/server/api/controllers/vessel-reports.controller.js
zkwsk: By the way: check this article to read about destructuring with default values: 

https://wesbos.com/destructuring-default-values/

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -3,13 +3,50 @@
 const HttpError = require("../lib/utils/http-error");
 const knex = require("../../config/db");
 
-//Get vessels reports by vessel id
-const getVesselsReportByVoyageId = async voyage_id => {
-  const vesselReports = await knex.from("vessel_reports").where({ voyage_id });
-  if (vesselReports.length === 0) {
-    return `No voyage exist with that id ${voyage_id}.`;
+//Get vessels reports by voyage_id with limit, offset=0 and order =desc
+const getVesselsReportByVoyageIdWithParams = async (voyage_id, query) => {
```
zkwsk:
```suggestion
const getVesselsReportByVoyageId = async (voyage_id, query) => {
```

Please keep the original name. We are not going to have a `getVesselsReportByVoyageIdWithoutParams` and overtime we will be adding pagination to most of the endpoints anyway, so not much point in changing the name.

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -3,13 +3,50 @@
...
+      .offset(0);
+
+    if (vesselReports.length === 0) {
+      throw new HttpError(
+        "Bad request",
+        `No vessel-reports exist with that voyage id ${voyage_id}.`,
+        404
+      );
+    }
+    return vesselReports;
```
zkwsk: Your logic does not allow me to any of the parameters together. If I do this, offset will be ignored:
```
http://localhost:3000/voyages/3/vessel-reports?limit=5&offset=2
```

If you first set default values for all of the possible query parameters, then do a single knex query with all the parameters set and without all of the if/else logic it should work.

zkwsk: Also makes it a lot simpler to implement on other endpoints later.

zkwsk (changes requested):
Some comments. Please remember to update swagger.json as well :)

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: resolved with help of Zaki.

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: removed

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: changed the method name and added method  swagger.json. Tested.

src/server/api/controllers/vessel-reports.controller.js
Swapna-Sahu: Working good. Thanks Zaki for recoding.

zkwsk (approved):

src/server/api/routes/voyages.router.js
```diff
@@ -45,13 +45,13 @@ router.post(
   }
 );
 
-//ENDPOINT: /api/voyages/:voyage_id/vessel-reports/
+//ENDPOINT: /api/voyages/:voyage_id/vessel-reports/?offset=0&limit=50&orderBy="desc"
```
ismailirahim, line 48: Limit and offset its working fine but i cant get the orderBy "desc" or "asc"!

src/server/api/controllers/vessel-reports.controller.js
```diff
@@ -3,12 +3,29 @@
...
-  const vesselReports = await knex.from("vessel_reports").where({ voyage_id });
-  if (vesselReports.length === 0) {
-    return `No voyage exist with that id ${voyage_id}.`;
+//Get vessels reports by voyage_id with limit, offset=0 and order =desc
+const getVesselsReportByVoyageId = async (voyage_id, query) => {
+  const voyage = await knex.from("voyages").where({ id: voyage_id });
+
+  const limit = parseInt(query.limit, 10) || 50;
+  const offset = parseInt(query.offset, 10) || 0;
+  const orderBy = query.orderBy || "desc";
```
ismailirahim, line 12: OrderBy its not working

PR #127, opened 2019-10-15
Run migrations on heroku

zkwsk:
So our heroku environments have been lying around stale because I did not find time to sit down and look into running migrations on deploy.

I finally managed to do so and found a few other things that needed fixing before being able to run things live:
- Made migrations run on deploy
- The way I had declared the api-path previously was super stupid. I did a bit of refactoring.
- Since running on Heroku will automatically set the `NODE_ENV=production` I discovered that we only had a "development" section in the knexfile, so I had to expand that with a production section. Later on we need to set it up so we will be running a separate set of seeds on production environments, so our seed data does not everwrite real data, but for now we only have seed data. But this can be done from the knexfiles production section.
- Express needed to be set up to take any arbitrary request and forward it to the react client application. Without this you would never be able to deeplink to a resource inside react.
- Added a favicon :)


Swapna-Sahu (approved):
🙌🏼

PR #135, opened 2019-10-17
Voyage details container

Swapna-Sahu:
Do not merge


src/client/components/VoyageDetailsContainer.js
```diff
@@ -1,5 +1,180 @@
...
+    hsfo: null,
+    ulsfo: null,
+    hfo: null,
+    lsfo: null,
+    hfocosts: null,
+    lsfocosts: null,
+    totalcost: null,
+    depart_from_port: null
+    //arrive_at_port: null
+  };
```
zkwsk, line 38: That is one big state object 😮 

But it's probably fine to do it like that. Excepts that I would expect you to have a full `suggested_route` object in there with the current selected suggested route, rather than just the id of that one. Would also make it easier to get the waypoints.

src/client/components/VoyageDetailsContainer.js
```diff
@@ -1,5 +1,180 @@
...
+
+  componentDidMount() {
+    const headerObject = {
+      // method: "GET",
+      headers: {
+        "Content-Type": "application/json",
+        authorization:
+          "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MSwicm9sZSI6InN1cGVydXNlciIsImlhdCI6MTU3MTM4MTI0NSwiZXhwIjoxNTcxNDE3MjQ1fQ.AWtUrRse7Q2_Z0FC0zcAqK1r2BR8BVBCaUz4Snc3Iy0"
```
zkwsk, line 46: Nice that you've set the header once for this file. Make sure to get the token from here, as soon as this PR is merged: https://github.com/HackYourFuture-CPH/ov-class09-fp/pull/137

src/client/components/VoyageDetailsContainer.js
```diff
@@ -1,5 +1,180 @@
...
+          ulsfo={lsfo}
+        />
+        <CostWidget
+          totalCost={totalcost}
+          hfoCost={hfocosts}
+          lsfoCost={lsfocosts}
+          hfoConsumption={hfo_comsumption}
+          lsfoConsumption={lsfo_comsumption}
+        />
+      </>
```
zkwsk, line 197: Once you put the map in there, we will have big problems because I renamed all of the props on the map and on the vessel list components to be camelCased. But if you want to avoid refactoring your stuff you could pass the props through the method I'm using to camelCase all of the keys before sending the props into Map.

You can use the package "camelcase-keys" to do so. See how I'm doing it here: https://github.com/HackYourFuture-CPH/ov-class09-fp/compare/vessel-route-on-map-zaki#diff-2106df1bbc7f38d27d9063944bf1a0bdR26 (remember that you don't have it in your npm dependencies yet)

src/client/components/VoyageDetailsContainer.js
Swapna-Sahu: I have the selected route table from there I am getting waypoints

PR #137, opened 2019-10-18
Token utility

zkwsk:
This allows you to easily get the `user_id`, `user_name`, `user_role` or `organization_id` stored in the authentication token. Just do:

```
import { getTokenData } from "../utilities/getTokenData"; // Make sure you use the right path according to the file you are importing it in

// ...
// YOUR CODE
// ...

const organizationId = getTokenData("organization_id");
```

To get the authentication token do the following:

```
import { getAuthToken } from "../utilities/getTokenData"; // Make sure you use the right path according to the file you are importing it in

// ...
// YOUR CODE
// ...

const authToken = getAuthToken();

// And use it with Axios for example like this
axios
  .get(`/api/organizations/${organizationId}/voyages?status=ongoing`, {
       headers: { authorization: getAuthToken() }
     })
     .then(response => {
     // Success
   })
  .catch(error => {
    // Fail
  });
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "generate-password": "^1.4.2",
     "interactjs": "^1.5.3",
     "jsonwebtoken": "^8.5.1",
+    "jwt-decode": "^2.2.0",
     "knex": "^0.19.0",
     "leaflet": "^1.5.1",
     "moment-timezone": "^0.5.26",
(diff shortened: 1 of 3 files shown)


NayanaKamtekar (approved):
It looks fine.....

PR #140, opened 2019-10-18
Voyage Teaser UI

ghost:
Note: instead of Button we can use React Link and also it needs to be styled.


src/client/components/VoyageTeaser.js
```diff
@@ -1,6 +1,37 @@
...
+//import material-ui
+import { makeStyles } from "@material-ui/core/styles";
+import Card from "@material-ui/core/Card";
+import CardActions from "@material-ui/core/CardActions";
+import CardContent from "@material-ui/core/CardContent";
+import Button from "@material-ui/core/Button";
+import Typography from "@material-ui/core/Typography";
+
+//CSS Colors: #F2F2F2,#D9D9D9,#595959,#262626,#0D0D0D
+const voyage_teaser = makeStyles({
```
AbodHassam: You can change it to voyageTeaserStyle or something like that

AbodHassam (changes requested):

src/client/components/VoyageTeaser.js
ghost: @AbodHassam First I need to be sure Zaki is accepting this style then I can change the name as well.

Swapna-Sahu (commented):
I think we are changing everything in camelCase.

src/client/components/VoyageTeaser.js
zkwsk: Agree. It should be camelCased.

I wanted to do something like this to simplify things:

```
const classes = makeStyles({ ...styles...})
```

But I did a quick test and it was breaking and I could not figure out why. Can you try if you can get it to work? Would be nice to kill that intermediary variable assignment (`classes = voyage_teaser()`).

src/client/components/VoyageTeaser.js
```diff
@@ -9,16 +41,33 @@ function VoyageTeaser({
...
+        </CardContent>
+        <CardContent className={classes.summery}>
+          <Typography className={classes.infoTitle}>ETA</Typography>
+          <Typography className={classes.describeText}>{arrivesAt}</Typography>
+          <Typography className={classes.describeText}>
+            {target_arrival_time}
+          </Typography>
+        </CardContent>
+        <CardActions>
+          <Button size="small">Show Details</Button>
```
zkwsk: Could you make the following changes to the show details button:

- [x] Make the font size 1.25rem
- [x] Align the button to the right
- [x] Make the font bold
- [ ] Use <Link> component.

zkwsk (changes requested):
Overall, can you please change the following:

- [x] Make the background of the entire page the color #EDF3F8 (Please do so in App.js)
- [x] Make the cards white
- [x] Apply margin-top, margin-left and margin-right on each of the cards to be 0.75rem
- [x] You must have set max-width of the cards somehow. I could not quite see it in your code, but if I remove your styling it's not there. Figure out how to remove that, so the cards will expand to the full width of the screen. We will put it all into a layout so the cards will be next to the map later.

src/client/components/VoyageTeaser.js
zkwsk: Oh, yeah. I agree that we should use a `<Link>` component here. Was trying to do that together with Nayana earlier though and got strange errors, but we were linking around the entire card. Now I only want the button linked, so please try that as well.

ghost: In my opinion, 1.25rem for the button is too much but you are the one who decides.
an I didn't understand what you mean by "Use Component", Do you want me to make a new Component for button and import it to the "VoyageTeaser"?

zkwsk (changes requested):
Sorry, I'm not sure what I tried to say with the "use component" comment. I think something got deleted and I did not see it. Maybe I asked you to add the `<Link/>` component.

Please do so. You must link to the respective voyage details page, so the link would be ``/voyages/${voyageId}``

Further, I wanted to tell you to change the text color on the button. It should be #0B3C61

Swapna-Sahu (dismissed):
Looks good to me. Good Job :)

Swapna-Sahu (approved):
Beautiful :)

zkwsk (approved):

PR #151, opened 2019-10-20
Create user with random password

AbodHassam:


ismailirahim (dismissed):
Looks good to me !

src/server/api/controllers/user.controller.js
```diff
@@ -83,7 +83,14 @@ const createUser = async ({ body }) => {
     });
 
   if (role.length > 0) {
-    const hashedPassword = await hashPassword(body.password);
+    var password = generatePassword.generate({
```
zkwsk:
```suggestion
    const password = generatePassword.generate({
```

zkwsk (changes requested):

zkwsk (approved):

PR #159, opened 2019-10-20
Added material-ui wrapper component for voyage map and list/details

NayanaKamtekar:
Added material-ui wrapper component for voyage map and list/details


zkwsk (changes requested):
As I just wrote you on slack I'm a little worried about introducing more layers of components for little things like this, because we may have to pass props through many layers. Using grid is something really basic we will do inside a lot of components anyway.

So could you just set up the grid in the render method of the 3 container components that have maps, instead?

Furthermore, could you set the css for leaflet (in `/public/index.html`) to the following:

```
    .leaflet-container {
        height: 100vh;
        width: 100%;
    }
```

We probably want to put that CSS somewhere else at some point, but not right now :)

zkwsk (approved):

PR #160, opened 2019-10-20
Secondary navigation bar

fawadik:
seondary navigation menu added to all components


src/client/App.js
```diff
@@ -6,6 +6,7 @@ import MainRouter from "./routes/main-router";
 import "./App.css";
 
 import NavBar from "./components/NavBar";
+import SecondaryNavigationBar from "./components/SecondaryNavigationBar";
```
lil2400, line 9: Remove SecondaryNavigationBar from Import.

src/client/index.js
```diff
@@ -7,7 +7,7 @@ import * as serviceWorker from "./serviceWorker";
 import axios from "axios";
 import { getAuthToken } from "./utilities/getTokenData";
 
-axios.defaults.headers.common["Authorization"] = getAuthToken();
+//axios.defaults.headers.common["Authorization"] = getAuthToken();
```
lil2400, line 10: Why you have commented this line?

lil2400 (changes requested):

src/client/index.js
zkwsk, line 10: Yeah. That's a problem. None of the API calls are authorized now.

zkwsk (changes requested):

zkwsk (changes requested):
There are a bunch of things breaking with this PR:

- [ ] Please do not comment out the global axios configuration. If you had problems you probably had not pulled develop into your branch.
- [ ] Do not insert the secondary navigation on each page. It needs to be placed inside of the NavBar component. Not under it.
- [ ] Please remove the button markup around Abods logout button. It is only used here, so you can do anything it takes to make it look properly with his component.
- [ ] All of the split screen layouts are broken in this PR.

Make sure to fix these things. Pull the latest develop and test it agian.

PR #161, opened 2019-10-20
Materialize suggested route cards

ghost:
Material UI is added to some of these components,


Swapna-Sahu (approved):
Everything works properly in storybook. Styling is bit mis-matching. Zaki will confirm styling if it is good.

PR #162, opened 2019-10-20
Material ui voyage details container

Swapna-Sahu:
Added Material-UI design for below files:
VesselDetails
CostWidgets
SuggestedRouteTable


JustinaOyov (commented):
Nicely wired UI. Loads fine in both Storybook and React app.

JustinaOyov (dismissed):
Nicely wired UI. Loads fine in both Storybook and React app.

src/client/components/VoyageDetails.js
```diff
@@ -16,30 +51,60 @@ function VoyageDetails({
   hsfo,
   ulsfo
 }) {
+  const classes = useStyles();
   return (
-    <div>
+    <Card className={classes.card}>
```
zkwsk: Please do not put a card around the entire component. The teaser already has its own card. The result is that we get multiple cards nested inside each other.

Instead wrap the section below it and the selected routes table.

Like this: 

Screen Shot 2019-10-21 at 11 20 52

zkwsk (changes requested):

src/client/components/VoyageDetails.js
Swapna-Sahu: Resolved.

zkwsk (approved):

PR #167, opened 2019-10-21
private route for voyages

hemagunti:
added private route for voyages


zkwsk (approved):
Beautiful! Thank you 🙏

